PR #22, opened 2017-02-21
update formatting of command line and link at end of sentence

ahus1:
For commands to be typed on the command line I recommend a "listing" block instead of an "example" block. This way line breaks are preserved and it looks more like a console. With the example block the current console example don't show at all on Github.

Also Asciidoctor known that "." just after a URL is not part of the URL. Therefore you don't need a blank there.


rdmueller:
Cool! Many thanx for the fixes!

PR #37, opened 2017-06-16
fix #34 set dependencies explicit.

### LINKED ISSUE #34: Deep nested structure
At the moment we have a really nested structure. I wondering if it wouldn't become more handy to have the following structure:

At top level
- src/doku and src/doc: based on the assumption that nobody writes a documentation in two languages and even if he does it's better to create than 2 git repos.
- under doku or doc only  arc42, images and perhaps pdfTheme (even this is not really documentation it's more styling) and than no further nesting. 

It's make it easier to delete all the stuff it's not necessary and easier to navigate.

opensource21:
Added the dependencies


opensource21:
Wrong comment

PR #38, opened 2017-06-16
fix #36 set dependencies explicit.

### LINKED ISSUE #36: Gradle doesn't detect all changes
If a change is only make at a subdocument, for example the glossar, gradle thought that the output is newer than the input and do nothing.

opensource21:


rdmueller:
Thanx! looks great!

PR #39, opened 2017-06-19
Fix #33 Add Excel export. 

### LINKED ISSUE #33: ExcelToCSV-Export
I wondering if you have a task which exports every sheet of excelfiles to csv, so that they can included into asciidoc. 
Can you provide an idea how to add a gradle task to the project? Must there a task for maven too?

opensource21:
Export all Excel-Sheet into a separate directory as csv-files.


rdmueller:
wow. that was quick. Thanx!
btw: if you would like to write a blog post about how you wrote this task - a PR to https://github.com/rdmueller/rdmueller.github.io/tree/master/_posts would be the right place :-)

rdmueller:
...and btw: if you would like to document your task:
https://github.com/rdmueller/docToolchain/tree/master/src/docs/manual :-)

PR #45, opened 2017-07-06
Feature: language type for confluence

MoePad:
Before, something like this was not processed correctly:

```
[source,bash]
----
some text
----
```

Now, the language type will not be ignored but correctly added to the confluence macro.

With the above example, you will get a correct code block in confluence using the bash highlighting.

To achieve this, several steps were necessary:
1. Sanitize the codeblock from added XML syntax (i.e. `<span class="type">`, etc.)
2. skip in the general replacement block all code tags, that contain a language attribute
3. replace all code tags with the correct syntax

For further question, please write me (we can use german for a private conversation)


rdmueller:
thanx for the PR!
Very much appreciated!

PR #46, opened 2017-07-06
make title of page containing preamble configurable

bodewig:
The text between the title and the first second level heading (the preamble) is put onto a separate page who's title is a hard-coded "arc42".

This PR allows users to override the page title for each input document.


rdmueller:
Thanx for this PR!

Issue #47, opened 2017-07-08
add an extra HTMTL snippet to each Confluence page


bodewig:
When publishing to Confluence we'd like to include a "Do not edit this page" warning macro to every page created.

I'll open a pull request that allows an optional extra content to be configured.

PR #48, opened 2017-07-08
allow optional extra content to be added to each published page

bodewig:
see #47


rdmueller:
thanx! I will use that feature too!

PR #52, opened 2017-07-09
promote headers in sect2 Confluence pages as well

### LINKED ISSUE #51: "headline promotion" in Confluence export is inconsistent
Right now when `confluenceCreateSubpages` is `true` all headlines get "promoted" to higher levels (i.e. `h4` becomes `h1` and so on). Shouldn't the same also be done for the pages created from `h2` as well, just with one level less of promoting - i.e. `h3` becomes `h1` and so on?

I'll be happy to take a stab myself, Before I do so I just want to make sure the difference is not an intentional one.

bodewig:
closes #51

### DIFF
--- a/scripts/asciidoc2confluence.groovy
+++ b/scripts/asciidoc2confluence.groovy
@@ -399,6 +399,12 @@ def recordPageAnchor = { head ->
     a
 }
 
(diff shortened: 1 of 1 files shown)


rdmueller:
thanx!
nice and groovy implementation!

Issue #53, opened 2017-07-10
StackOverflow in asciidoc2confluence in script-Tags regex


bodewig:
There seems to be a problem with the first regexp introduced with #45 - it causes a stack overflow

```
java.lang.StackOverflowError
	at java.util.regex.Pattern$Neg.match(Pattern.java:5086)
	at java.util.regex.Pattern$Branch.match(Pattern.java:4604)
	at java.util.regex.Pattern$GroupHead.match(Pattern.java:4658)
	at java.util.regex.Pattern$Loop.match(Pattern.java:4785)
	at java.util.regex.Pattern$GroupTail.match(Pattern.java:4717)
	at java.util.regex.Pattern$BranchConn.match(Pattern.java:4568)
	at java.util.regex.Pattern$CharProperty.match(Pattern.java:3777)
	at java.util.regex.Pattern$Neg.match(Pattern.java:5097)
	at java.util.regex.Pattern$Branch.match(Pattern.java:4604)
...
```

with all lines I can get stemming from ` java.util.regex.Pattern` - it is this one https://github.com/rdmueller/docToolchain/pull/45/files#diff-8e03331c0e61a64770510b8e9d5a7021R205

I'm trying to create a minimal document triggering the problem. It looks as if one needs two code blocks with language attributes on the same Confluence page. First suspect is the negative look ahead and I need to figure out why it is there before I try to fix it :-)

It might be possible to achieve the same effect by using JSoup rather than regexps.

bodewig:
I was wrong, you only need a single code block, at least when using the `coderay` syntax highlighter. I'll attach an example that causes a stack overflow.

bodewig:
https://github.com/bodewig/docToolchain/tree/fix_code_block_stackoverflow is close to a fix AFAICT, only the unescaping of HTML isn't working. I'll be getting there :-)

bodewig:
I've added another commit and it works for the files I've tested, am waiting for confirmation by a co-worker before I create a pull request.

@MoePad it would be great if you could verify I haven't broken anything for you with https://github.com/bodewig/docToolchain/tree/fix_code_block_stackoverflow

@rdmueller you mentioned invalid XHTML in #44 - any chance you've still got some of these files around in order to verify they get published properly after applying my changes? I've removed two of the `replaceAll` lines you seem to have added because of them.

MoePad:
If I can make it, I can have a look at that this evening. Otherwise my next available time slot will be Thursday.

[...]

PR #55, opened 2017-07-11
Fix code block stackoverflow

bodewig:
see #53 

This works for us, but there may be some cases that the changes miss, see the discussion inside the issue itself.


rdmueller:
Wow. ich komme ja kaum noch hinterher :-)

bodewig:
Let's see what @MoePad says in #53 - I hope merging has not been premature.

Many thanks!

MoePad:
I've just tested it for java and bash code blocks: it works fine! Good work! :)

bodewig:
Thanks!

PR #50, opened 2017-07-08
make internal cross references work for Confluence pages

bodewig:
For the longish story see #49 

I've stripped all HTML markup from link texts now, seems to work well for our use case.

PR #56, opened 2017-07-11
overlooked one case, where realTitle should be used

bodewig:
in #50 I introduced the `realTitle` closure and overlooked one place where it should be used


rdmueller:
perfect. thanx.

PR #58, opened 2017-07-12
[BUGFIX] Reference Online Images (Confluence)

MoePad:
Currently, it was not possible to reference online images (i.e. 'http://server/image'). The confluence script failed with an error.

This fix will check if the image source starts with 'http' and skip the upload to confluence and instead use another tag to correctly include online images.

Please test before merging, because my setup currently only includes one local image (which worked fine).

### DIFF
--- a/scripts/asciidoc2confluence.groovy
+++ b/scripts/asciidoc2confluence.groovy
@@ -252,13 +252,21 @@ def parseBody =  { body, anchors, pageAnchors ->
         }
         def src = img.attr('src')
         println "    image: "+src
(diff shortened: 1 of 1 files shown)


MoePad:
The changes are also here: https://github.com/MoePad/docToolchain.git branch bugfix-reference-online-images

bodewig:
I've successfully tested it with a few documents as well, but none of them has any external image references.

MoePad:
Sounds good ;)  I am more worried about the local images than the external ones (i.e. that nothing is broken that already worked).

rdmueller:
great! so let's merge it...

PR #59, opened 2017-07-12
sort child pages by creation in children macro

### LINKED ISSUE #54: Confluence "Children Macro" Sort Order
When exporting to Confluence the "children" macro is added at the bottom of each page. The script doesn't specify any order explicitly and it seems "sort by title" is the default choice.

When using section numbers this starts to look weird as soon as you get to chapter 10 :-)

The other sort options are "creation" and "modification".

Given the way the script works, pages should get created in the same order they appear inside the document that is being exported. So maybe explicitly sorting by "creation" would be better?

The alternatives are "leave it as it is" and "make it configurable".

bodewig:
closes #54


rdmueller:
thanx!

Issue #25, opened 2017-05-02
exportToConfluence: Definition-Lists


rdmueller:
a definition list like this:

```
Definition::

a test
```

results in a malformed html:

`Error parsing xhtml: Unexpected close tag </table>; expected </tr>.`

```
<div class="dlist"> 
 <table><tr> 
  <tr><th>
   Definition
  </th> 
  <td> 
   <p>a test</p> 
  </td></tr> 
</table> 
</div> 
```

bodewig:
Doesn't seem to happen for me, maybe this has been fixed as a side effect by now? Or maybe it depends on the version of asciidoctor (am using 1.5.5).

rdmueller:
great! so let's close this...

bodewig:
In general definition or description lists are a bit more flexible than pairs of `dt`s and `dd`s, you can have multiple elements on either side in order to share a description for more than one term or have multiple descriptions (in different languages, for example) for one term.

I don't think asciidoctor creates `dl`s with n:1 or 1:n structures right now but your example looks as if you've had a `dt` without a `dd` at one point in time.

I've tried to make the `dl` handling more robust with https://github.com/rdmueller/docToolchain/compare/master...bodewig:more_robust_dl_handling - which is certainly a lot more code than the current `replaceAll` approach.

If you think the extra effort is worth it, I'll be happy to turn my branch into a PR. If you feel it isn't needed that's no problem, either :-)

rdmueller:
I've only implemented features which I used.
So, yes, definition lists could be implemented better and I am happy to accept all PRs which make docToolchain better!

bodewig:
Here you go #60 :-)

PR #60, opened 2017-07-15
deal with descriptions lists with unbalanced dt/dd elements

bodewig:
This might prevent problems like #25 in the future.


bodewig:
Thank you for creating it in the first place.

PR #61, opened 2017-08-24
Adopted image attributes _width_ and _align_ to confluence

MoePad:
The current implementation does not respect the custom configuration of an image (i.e. the image width and the image alignment).
Those two attributes are parsed from the original and passed to confluence.


rdmueller:
Very cool - thanx!

Issue #74, opened 2017-09-13
shift more properties to the gradle.properties file


rdmueller:
the gradle.properties file should be the standard for properties definitions since it can be overwritten by the command line:

https://docs.gradle.org/current/userguide/build_environment.html#sec:gradle_properties_and_system_properties

rdmueller:
thanx for the PR.
I think we can close this for the remaining properties as duplicate of #310

PR #75, opened 2017-09-14
Add confluenceConfigFile to gradle.properties

jakubjab:
As a first step for #74, I made the path to the confluence configuration file available for change from the command line:

`./gradlew -PconfluenceConfigFile=$HOME/ConfluenceConfig.groovy publishToConfluence`

I didn't test that on Windows.


scripts/asciidoc2confluence.groovy
```diff
@@ -57,13 +57,8 @@ def baseUrl
 
 // configuration
 def config
-try {
-    println "scriptBasePath: ${scriptBasePath}"
```
jakubjab: I don't know in which case `scriptBasePath` would be available if run from build.gradle, please clarify if that's needed for other reasons.

rdmueller: afaik, the scriptBasePath was introduced for the stand-alone version of asciidoc2confluence

rdmueller:
first of all: sorry, but I created a merge conflict through another PR

as I can see, your change makes sense but I wonder if we could even make it more powerfull:
what if we manage to move all the whole `ConfluenceConfig` to the `gradle.properties`?

The problem I do see is that the `gradle.properties` are not dynamic - they can't contain groovy code. So we would have to move this functionality (like asking for a password if password is empty) to the core script.

I think we can also move the asciidoctor2confluence script back to the gradle build - this way we would not have to `eval` it. I think I once moved it to a script in order to have it more modular but now with the script plugins, I don't see a need for that anymore.

what do you think?

jakubjab:
I will rebase my changes, no problem.

As for configuration in general, there are some aspects here:
1. I would like that all configuration is done in the same way, so that it's easy to use
2. gradle.properties files don't allow data structures more complex than a string, while it may be useful to have a list of strings or a map, e.g. list of input files
3. I cannot store passwords in project's gradle.properties

It would be indeed much simpler if we could agree on gradle.properties for everything. Passwords can be passed through `~/.gradle/gradle.properties`, and having only single input file is fine for me.

rdmueller:
maybe we can shift everything to a yaml based configuration?
https://plugins.gradle.org/plugin/pl.softmate.gradle-properties-yaml-plugin

rdmueller:
thanx for rebasing!

jakubjab:
About yaml configuration - I'm not sure if version 0.0.2 is stable enough. On the page it says:

> Please note, that current implementation is more PoC than production-ready code.

Personally, I need something stable and configurable, even if it had less features. I would stick with gradle.properties, as it seems the simplest possible solution.
Then as an extension we could add a path to configuration file, either yaml or groovy, which could be used by more demanding users.

rdmueller:
that sounds good. yaml for advanced users and properties as stable feature.

but to be honest, parsing yaml in groovy is a one-liner. A bit harder is to check the different locations in which a yaml file could be found. Maybe we can push the plugin to a stable 1.0 over time... I'll take a look at it...

bodewig:
For our use-case having a separate `asciidoctor2confluence` was great, as this is all that we needed. But I agree that treating it the same way as all the other plugins is beneficial in the long term.

Not being able to use groovy code inside of the configuration may be limiting for advanced use cases, though. It would be good to provide an alternative for it and I'm not sure the yaml plugin (or the properties file based https://github.com/stevesaliman/gradle-properties-plugin ) would be sufficient.

For example we have this

```
extraPageContent = System.getenv()['PAGE_BANNER'] ? System.getenv()['PAGE_BANNER'] :
    "<ac:structured-macro ac:name=\"warning\" ac:schema-version=\"1\"><ac:parameter ac:name=\"title\" /><ac:rich-text-body>This is a read-only copy of the concept developed at <a class=\"bare\" href=\"${System.getenv()['BANNER_URL']}\">${System.getenv()['BANNER_URL']}</a>. Please do not edit this page directly as all your changes will be overridden with the next update.</ac:rich-text-body></ac:structured-macro>"
```

while I can get the effect of environment variables with properties files I won't get the logic part.

TBH I'm not sure our use case is representative, though, and we could probably find a way to make things work with some creative groovy gymnastics for us even with properties files.

rdmueller:
@bodewig - thanx for the valuable input. You example shows the power of the groovy configuration files and I too would like to find a solution which does not remove this feature.
let's approach this step by step

bodewig:
Just in case it was unclear (happens all the time), I really didn't mean to imply you shouldn't move to properties. And I'm all for incremental improvements :-)

Many thanks.

jakubjab:
I'll think about it and try to work on that soon.

Issue #71, opened 2017-09-11
Separate tool from data


jakubjab:
I would like to have docToolchain separately as a tool, which can be imported into our build system and used for multiple software systems without changing.

Currently src directory is embedded in the tool, which makes it hard to maintain - we cannot just pull new version of the tool from GitHub and use it without moditications.

It's also licensing problem, because the license for the tool may be different than the license for the documentation itself.

My proposal is to pass all the parameters to gradlew in the command line (e.g. with -P option), so that we could run e.g. gradlew -PsourceDir=/someProjectA, etc.

I started to work on that, but before continuing I'd like to agree with authors.

rdmueller:
sounds like a plan!
I also noticed that within the repo, the tool, manual, arc42 template and examples are mixed. In addition, I would like to add some automated tests.

So a better structure would be a great idea - can you suggest one?

And please take a look at the deck.js branch. There I have already "modularized" the build file. It now consists mainly of references to script plugins. Maybe that's a start...

@opensource21 , @bodewig : what are your thoughts?

bodewig:
I'm all for the separation. For our local projects I'm maintaining a local fork of the tooling (and only the tooling) as this is all we need.

rdmueller:
So, I wander what will be the best way to make this modular.
We have three big modules:
- docToolchain itself with or without arc42
- the eat-your-own-dogfood manual with examples
- hopefully some more tests in the future

1) I personally think it would be cool to have something like http://start.spring.io/ to create your own flavor.

2) docToolchain could also be divided into several Gradle plugins and you manually select your tools you need.

3) third option is to break it all up into three repositories.

What do you think? Which option do you prefer? Are there even additional options?

@jakubjab : about the licensing: do you already see a licensing problem?

jakubjab:
I would vote for splitting it into separate repositories:
1) docToolchain without any template
2) manual and examples
3) arc42 template

If you want just to use the tool, download 1) and include in your build system.
If you want to use and develop the tool, download 1) and 2).
If you want arc42 template, download 3), or use any other template available. You could consider if it's not better just to refer to official arc42 template instead of maintaining your copy, see https://github.com/arc42/arc42-template.

As for licensing, current docToolchain includes arc42, which is also licensed separately on Creative-Commons Sharealike 4.0., see http://faq.arc42.org/questions/A-2/. That's a little unclear, but probably not a big problem.

rdmueller:
[...]

PR #85, opened 2017-09-22
Add custom configuration.

jakubjab:
First attempt at #71, tested with generating HTML and PDF on MacOS.
I'm new to gradle and groovy, so feel free to comment.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,21 +11,22 @@ plugins {
 }
 println "compiling with Java ${System.getProperty("java.version")} at ${new Date().format("HH:mm:ss")}"
 
+// configuration
+def config
+println "mainConfigFile: ${mainConfigFile}"
+config = new ConfigSlurper().parse(new File(mainConfigFile).text)
+println "inputPath: ${config.inputPath}"
+println "inputFiles: ${config.inputFiles}"
+println "outputPath: ${config.outputPath}"
+
 ext {
-    srcDir  = "$projectDir/src/docs"
-    targetDir = "$buildDir"
+    srcDir  = config.inputPath
+    targetDir = config.outputPath
     javaVersion = System.getProperty("java.version")
     currentDate = new Date().format("d. MMM yyyy")
     // where HTMLSanityCheck checking results ares stored
-    checkingResultsPath = "$buildDir/report/htmlchecks"
-    sourceFiles = [
-                    [file: 'arc42-template-de.adoc', formats: ['html','pdf','docbook']],
-                    [file: 'arc42-template-en.adoc', formats: ['html','pdf','docbook']],
-                    [file: 'arc42-template-es.adoc', formats: ['html','pdf','docbook']],
-                    [file: 'test.adoc',              formats: ['html','pdf','docbook']],
-                    [file: 'manual.adoc',            formats: ['html']],
-                    [file: 'ppt/Demo.pptx.ad',       formats: ['html','revealjs']],
-                ]
+    checkingResultsPath = "${config.outputPath}/report/htmlchecks"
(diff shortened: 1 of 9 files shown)


rdmueller:
very cool! You just tested our first automated travis tests!
I'll review the changes later today before I merge...

build.gradle
```diff
@@ -156,7 +156,7 @@ task generatePDF (
         description: 'use pdf as asciidoc backend') {
 
     attributes \
-        'plantUMLDir'         : file('build/docs/images/plantUML/').path
+        'plantUMLDir'         : file("${config.outputPath}/images/plantUML/").path
```
rdmueller: I'm not sure if this is correct. The plantUMLDir is only used for PDF-generation. I had to introduce this because otherwise the PDF generator didn't pick the generated images up.
I will test this asap.

build.gradle
```diff
@@ -79,15 +80,13 @@ apply from: 'scripts/htmlSanityCheck.gradle'
...
-    inputs.file new File(srcDir, 'arc42-template-es.adoc')
-    inputs.file new File(srcDir, 'config-es.adoc')
-    inputs.file new File(new File(targetDir, 'docs'), 'changelog.adoc')
+  
+    config.taskInputsDirs.each {
+        inputs.dir new File(it)
+    }
+    config.taskInputsFiles.each {
+        inputs.file new File(it)
+    }
```
rdmueller: that's a great change! cleans up the dependencies...

build.gradle
```diff
@@ -16,21 +16,22 @@ configurations.all {
...
 println "compiling with Java ${System.getProperty("java.version")} at ${new Date().format("HH:mm:ss")}"
 
+// configuration
+def config
+println "mainConfigFile: ${mainConfigFile}"
+config = new ConfigSlurper().parse(new File(mainConfigFile).text)
+println "inputPath: ${config.inputPath}"
+println "inputFiles: ${config.inputFiles}"
+println "outputPath: ${config.outputPath}"
+
```
rdmueller: great! this gives more transparency...

rdmueller:
btw: thanx for also updating the manual!

src/docs/manual/02_install.adoc
```diff
@@ -78,3 +80,15 @@ The first build generated some file within the `<project>/build` folder:
...
 If you didn't get the right output, please raise an issue on https://github.com/rdmueller/docToolchain/issues[github]
 
+== Custom configuration
+
+If you want to use the same docToolchain for multiple projects, you need to create a copy of Config.gradle and ConfluenceConfig.gradle for each project.
+The paths to those configuration files can be provided by giving -P option to gradlew, for example:
+
+[source, bash]
+----
+./gradlew -PconfluenceConfigFile=$HOME/projectA/ConfluenceConfig.gradle -PmainConfigFile=$HOME/projectA/Config.gradle
```
rdmueller: hm. I guess these should be .groovy files, shouldn't they?
and shouldn't we name it 'config' starting with a lower case?

build.gradle
rdmueller: just wrote a test for it - plantUML still works. Will push my test as soon as it is cleaned up.

src/docs/manual/02_install.adoc
jakubjab: I created a new pull request fixing mistake in file names, they should be gradle of course.
If they should be upper or lower case, I don't have opinion, just existing ConfluenceConfig.groovy was upper case.

rdmueller: ah - yeah. I guess I should also rename this to confluenceConfig....

PR #87, opened 2017-09-25
Fix failing test for PDF generation with PlantUML

jakubjab:
I'm not sure how it worked for you, but it was failing for me as well as failed in CI without that change.
That $ ends up in the file path, which apparently confuses Linux or Mac, may be working on Windows.

### DIFF
--- a/src/test/docs/test.adoc
+++ b/src/test/docs/test.adoc
@@ -1,6 +1,6 @@
 include::config-de.adoc[]
 
-[plantuml,"${plantUMLDir}testPlant",png]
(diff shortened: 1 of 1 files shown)


rdmueller:
that funny - It really works on windows.
Ah! Just because it produces another error:

`asciidoctor-diagram: ERROR: Failed to generate image: Unknown error (SystemCallError) - Unknown Error (20047)
`
thanx for fixing it!

rdmueller:
btw: have you seen the slack channel? 
https://twitter.com/docToolchain/status/911994657676173314

PR #89, opened 2017-09-28
Add support for configuration of PDF styles

### LINKED ISSUE #88: PDF style per project
## Expected behavior and actual behavior.

Currently there is a single PDF style hardcoded in build.gradle, no way to configure it at runtime.
I would like to make PDF styles configurable per project, so that it's part of the contents, instead of being part of the docToolchain tool (which would remain as default).
See also #70 which is related.

## Steps to reproduce the problem.

N/A

## Specifications like the version of the project, operating system, or hardware.

N/A

jakubjab:
Define new "docDir" property, which serves as a base for all other
configuration paths.
Support for PDF styles included with the contents, not with the tool.
Fix conversion to epub and docx.
Fix minor faults with paths.

This is meant to fix #88, I can now specify custom PDF style with my documents.

### DIFF
--- a/Config.groovy
+++ b/Config.groovy
@@ -1,12 +1,14 @@
 
-// To use environment variables, you can use:
-// outputPath = "${System.getenv('HOME')}/docs"
-//
-// To specify paths relative to current directory, use PWD variable:
-// outputPath = "${System.getenv('PWD')}/docs"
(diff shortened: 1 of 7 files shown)


jakubjab:
Looks like Travis is not running here.

Issue #91, opened 2017-10-07
clean up and separate projects


rdmueller:
currently, the arc42 template and reveal.js project are copied into this project. 
It works, but is not a good style.
We should find a way to separate them again.

jakubjab:
@rdmueller - I was thinking of taking up this issue. How do you think this should be solved?
a) include reveal.js as submodule
b) require the user to get reveal.js themselves (possibly adding a script to make it easier)
The same for arc42.
My opinion is that reveal.js is part of the tool, so it could be a submodule, but arc42 is just a template, which could be recommended in the documentation, but shouldn't be hardcoded in the tool.

rdmueller:
Great!
a) I'm not sure what you mean by submodule? a git submodule? could make sense. But I guess it is important to fetch the right version. Great would be a solution where there is a build step on which the generateDeck step depends and which makes sure that reveal.js is available / downloaded. The download could be a real download or some kind of git statement (what do we need for a submodule? just a fetch?)
b) it should be as easy as possible for the user to use it. 

regarding arc42 - I still dream of some kind of init task. This could ask if the user wants to start with a template and offer some choices for a template. Currently, these will be the different arc42 templates, but could also be self-build ones or other publicly available ones.

So I think we have the same thoughts about the goal... :-) We just have to evolve the code...

jakubjab:
I'll give it a try, I'll let you know.

jakubjab:
I'm thinking how to remove custom version of arc42 template, and there are some differences comparing to https://github.com/arc42/arc42-template:
- changelog - Do we want to keep the changelog? Is that important? Do we want a changelog of the master template instead? Personally I would remove it, I don't see the need.
- open issues - Do you think it's important to list open issues as part of arc42? Could we fit that in chapter 11 of arc42 instead of creating an appendix?

rdmueller:
both changes are mainly examples of the changelog and issues feature.
So maybe we just enrich the docs about them and remove them from the custorm arc42 template

rdmueller:
I think we've did some progress on this by now and it is time to close this ticket.
Please open new issues if you think there is something missing.

PR #92, opened 2017-10-15
Move reveal.js and asciidoctor-reveal.js to submodules

jakubjab:
First part of #91, making reveal.js and asciidoctor-reveal.js submodules.
I have kept exactly the same versions as before, leaving upgrading out of this change.
The size of the repository is decreased from 13M to 7.2M.
The installation manual has become somehow more complicated, any suggestions are welcome. For example we could make init.sh script that would download all dependencies and move them to correct places.


jakubjab:
Thinking about it some more, we could also use something like https://github.com/craigburke/client-dependencies-gradle, so that Gradle would automatically download dependencies using bower.

rdmueller:
regarding bower: I would like to keep the dependencies low in order to get it easily up and running in a very restricted environment. So I prefere the submodule...

rdmueller:
maybe we could add the bower variant as an option?

rdmueller:
thanx for th PR and for updating the docs!

Issue #15, opened 2016-11-23
Change of character case in headline breaks publishToConfluence


rdmueller:
If the case of a character in a headline is changed, the script searches for the new headline and doesn't find a page with this name (case sensitive search).
It then tries to create the corresponding page which leads to an "page already exists" error (case insensitive)

Solution: make the search for the page title case insensitive

rdmueller:
see https://answers.atlassian.com/questions/44350209/answers/44916450/comments/44922027 for a workaround

rdmueller:
it might be that this is fixed by now.

For hacktoberfest, I would be happy if someone could verify this and add a remark to the docs (as PR) that headlines are now case insensitive

rdmueller:
was fixed by new local page search

PR #94, opened 2017-10-16
Fix issue #15 - change of character case in headline of Confluence

jakubjab:
publishToConfluence task is now searching for all pages within the space matching fuzzily the expected title. Then the page with the title matching the expected one is chosen in case-insensitive way. Pages with similar but different titles are not affected, even if matched by fuzzy search from Confluence.

### DIFF
--- a/scripts/asciidoc2confluence.groovy
+++ b/scripts/asciidoc2confluence.groovy
@@ -335,6 +335,25 @@ def parseBody =  { body, anchors, pageAnchors ->
     return pageString
 }
 
+def getPageByTitleIgnoreCase = { pages, realPageTitle ->
+    if (pages.size() >= 1) {
+        def page
+        pages.find { p ->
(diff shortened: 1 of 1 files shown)


jakubjab:
@rdmueller - any comments on that?

rdmueller:
sorry and thanx for the reminder! Will take a look at it this morning...

rdmueller:
nice code. As I do understand, it does a fuzzy search and looks in the result for the page which matches the original page case insensitive. Great!
Please see my comment and decide if we want to shorten the code...

rdmueller:
strange. You should be able to see it on this page...:

I guess this method can be replaced by

`def page = pages.find { page -> p.title.equalsIgnoreCase(realPageTitle) }`

and if it can be turned into a one-liner, I guess it can be written inline.

What do you think?

rdmueller:
cool. thanx.
are you registered at https://hacktoberfest.digitalocean.com/ ?

PR #100, opened 2017-10-23
Fix #57 by avoiding hijacking Confluence pages alredy existing

### LINKED ISSUE #57: relying on title in Confluence is dangerous
We just had a case where exporting one document "stole" a page from a different document because they happened to share the same title. We now made using `confluencePagePrefix` mandatory for us. The real problem is, we only noticed the problem by chance.

When the `parentId` is set `pushToConfluence` could verify the ancestor of the page it finds for a given title matches the `parentId` and either reject the exporting the page or at least print a warning.

jakubjab:
This solution is throwing an exception, which leaves the page possible half-updated, but I think it's better then "stealing" pages and overriding them.

### DIFF
--- a/scripts/asciidoc2confluence.groovy
+++ b/scripts/asciidoc2confluence.groovy
@@ -373,22 +373,15 @@ def pushToConfluence = { pageTitle, pageBody, parentId, anchors, pageAnchors ->
     }
 
     def pages
-    def foreignPages
     trythis {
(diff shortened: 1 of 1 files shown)


rdmueller:
looks fine to me when I browse through the changed code.
But I guess we need to establish some tests for the confluence code.
We have access to a confluence on demand - maybe we can use this for testing...
https://arc42-template.atlassian.net/wiki
I'll create an issue for that and try to come up with a solution

PR #124, opened 2017-11-21
Remove embedded arc42 template and make docToolchain standalone tool

jakubjab:
This is big one, so let's agree if that is the right direction, and then we can add tests to cover new functions.

This change removes the arc42 template from docToolchain.
New tasks to optionally download arc42 template are provided.
Documentation is updated accordingly.

### All Submissions:

### New Feature Submissions:

2. [x] Does your submission pass all tests? (see travis check)

### Change to Documentation:

* [x] Did you build the docs and copy them to the `/docs` folder?


jakubjab:
@rdmueller - any comments on that?

rdmueller:
@jakubjab the PR is great - thanx!
just gimme some extra time to check the changes

init.gradle
```diff
@@ -0,0 +1,93 @@
...
+) doLast {
+    copy {
+        from "${projectDir}/template_config"
+        into docDir
+    }
+}
+//end::initExisting[]
+
+//tag::initArc42DE[]
+task initArc42DE(
```
rdmueller: I think we should provide a parametrized task or maybe ask the user on the command line.
But we can enhance this at a later stage

scripts/exportChangelog.gradle
```diff
@@ -4,7 +4,7 @@ task exportChangeLog(
         group: 'docToolchain'
 ) doLast {
     // https://git-scm.com/docs/pretty-formats
-    def changes = "git log --pretty=format:%x7c%x20%ad%x20%n%x7c%x20%an%x20%n%x7c%x20%s%x20%n --date=short ./src/docs/arc42".execute().text
+    def changes = "git log --pretty=format:%x7c%x20%ad%x20%n%x7c%x20%an%x20%n%x7c%x20%s%x20%n --date=short ./resources/reveal.js".execute().text
```
rdmueller: hm. I think this should point to `./src/docs`, shouldn't it?
It is the folder for which the changelog should be fetched...

src/docs/manual/02_install.adoc
```diff
@@ -55,66 +53,100 @@ After unzipping, put the dependencies in `resources` folder, so that the structu
...
+./gradlew -b init.gradle initArc42EN -PdocDir=<your directory>
 ----
 
-.Windows with gradle wrapper
+The Config.groovy file is then preconfigured to use the downloaded template.
+
+== Build
+
+This should already be enough to start a first build. Due to some limitations in gradle, you need to run the build from
+the docToolchain home directory, and point docDir property to an absolute directory where your documents live.
```
rdmueller: is this really the case? have to check this...

src/docs/manual/02_install.adoc
```diff
@@ -55,66 +53,100 @@ After unzipping, put the dependencies in `resources` folder, so that the structu
...
 
-You can install docToolchain in a directory separate from your project. Then docToolchain is a standalone tool that can be
-upgraded independently from your code or used for multiple projects.
-If you want to install docToolchain separately, you need to create a copy of Config.groovy and
-ConfluenceConfig.groovy for each project. Inside each configuration file you need to specify the input and output
-folders and files.
+In addition to `Config.groovy` there is also a `scripts/ConfluenceConfig.groovy` file. If you are not using Confluence
+you can remove it. If you use Confluence, then you need to open this file and adapt to your environment.
+You can also create multiple copies of that file. For example you can have `ConfluenceConfig.groovy` for publishing
+official pages, and `MyConfluenceConfig.groovy` with a different Confluence space for reviews.
```
rdmueller: what happened to the `Custom configuration` section?

rdmueller:
let's merge this and check it out...

scripts/exportChangelog.gradle
jakubjab: Could be, but src/docs/arc42 doesn't exist any more, and other directories can change often. I have chosen reveal.js as external directory, thus having quite stable changelog. Otherwise the tests would need to be adapted after each change.

src/docs/manual/02_install.adoc
jakubjab: It's the default configuration now - it's up to the user if they wanted to use directory inside docToolchain, but honestly I don't see why they would. Then each upgrade of docToolchain gets more complicated.

src/docs/manual/02_install.adoc
jakubjab: I would be happy to avoid the need to cd to docToolchain, I would love to run this tool from the docDir. Let me know if you find any way.

scripts/exportChangelog.gradle
rdmueller: yeah, I will open an issue. We have to move this to the config.
It needs to be the folder for the documentation...

Issue #145, opened 2018-03-01
move AsciiDoc tasks to Script Plugin


rdmueller:
it would be nice to strip down the main build.gradle file to make it easier to merge it with your own project.
This could be achieved by moving the remaining asciidoc tasks to a script plugin

rdmueller:
solved

PR #151, opened 2018-03-03
#145 moved asciidoc tasks to script plugin

rdmueller:
This PR cleans up the main build file which makes it easier to incorporate docToolchain directly with existing projects.

### All Submissions:

* [X] Does your PR affect the documentation?
* [X] If yes, did you update the documentation or create an issue for updating it?

yes, docs should be ok. Even fixed some other issues.

Created another issue for creating some tutorials.

No, because there is no new functionality. But the ones still run fine.

2. [X] Does your submission pass all tests? (see travis check)

### Changes to Core Features:

* [ ] Have you written new tests for your core changes, as applicable?
* [X] Have you successfully ran tests with your changes locally?

### Change to Documentation:

* [X] Did you build the docs and copy them to the `/docs` folder?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,49 +2,15 @@
  * This build file is part of the docToolchain
  */
 
-import org.asciidoctor.gradle.AsciidoctorTask
-
 plugins {
     id "org.asciidoctor.convert" version "1.5.3"
     id "org.aim42.htmlSanityCheck" version "0.9.7"
     id "com.github.ben-manes.versions" version "0.15.0"
 }
-configurations.all {
-    resolutionStrategy {
-        force 'org.codehaus.groovy:groovy-all:2.4.12'
-    }
-}
-println "compiling with Java ${System.getProperty("java.version")} at ${new Date().format("HH:mm:ss")}"
-
-// configuration
-def config
-println "docDir: ${docDir}"
-println "mainConfigFile: ${mainConfigFile}"
-config = new ConfigSlurper().parse(new File(docDir, mainConfigFile).text)
-println "inputPath: ${config.inputPath}"
-println "inputFiles: ${config.inputFiles}"
-println "outputPath: ${config.outputPath}"
-
-ext {
-    srcDir  = "${docDir}/${config.inputPath}"
-    targetDir = "${docDir}/${config.outputPath}"
-    javaVersion = System.getProperty("java.version")
-    currentDate = new Date().format("d. MMM yyyy")
-    // where HTMLSanityCheck checking results ares stored
-    checkingResultsPath = "${docDir}/${config.outputPath}/report/htmlchecks"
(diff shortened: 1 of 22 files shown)


rdmueller:
Hi @jakubjab - this is the second PR I was talking about.
This moves the remaining task definitions to a script plugin.
Why is this so important for me?
If you have an existing Gradle project and want to add docToolchain functionality, you could either install docToolchain as stand-alone tool as defined by your shell script. But you could also add it to your build - this way, everybody with access to the source has also docToolchain.

Now with this change, it will be easier to copy all those script plugins instead of merging two build files.

It would be great to have this PR reviewd before the 14th of may (I will give a talk in madrid), but for me it is more important to get your thoughts and feedback.

Best Regards,
Ralf

build.gradle
```diff
@@ -55,202 +21,21 @@ dependencies {
...
 }
 
-asciidoctorj {
- version = '1.5.4'
-}
-
 
+//all available docToolchain modules
+//the first is mandatory
+apply from: 'scripts/AsciiDocBasics.gradle'
```
jakubjab: I would split out all generateX tasks to separate .gradle files as well, so that if a project doesn't need e.g. PDF then it's not bloated with unnecessary stuff.

docs/index.html
```diff
@@ -3082,9 +3097,9 @@ <h4 id="_q_i_get_an_error_message_saying_that_a_library_is_not_registered_when_i
 </div>
 </div>
 <div class="sect2">
-<h3 id="_sparx_enterprise_architect">3.3. Sparx Enterprise Architect</h3>
+<h3 id="_sparx_enterprise_architect">4.3. Sparx Enterprise Architect</h3>
 <div class="sect3">
-<h4 id="_q_sparx_enterprise_architect_is_a_windows_tool_isn_t_it">3.3.1. Q: Sparx Enterprise Architect is a windows tool, isn&#8217;t it?</h4>
+<h4 id="_q_sparx_enterprise_architect_is_a_windows_tool_isn_t_it">4.3.1. Q: Sparx Enterprise Architect is a windows tool, isn&#8217;t it?</h4>
```
jakubjab: This file needs to be re-generated, as it's missing latest fixes.
I was thinking to make a task to copy manual.html to index.html somehow automatically, otherwise it's easy to forget.

build.gradle
rdmueller: I thought about this too, but it seems that it's not easy to split the AsciiDoc tasks because of the 
task.withType(asciidoctor)
statement. I even hat to go back to the old plugin mechanism in order to get it this far...

docs/index.html
rdmueller: I once thought that Travis could do that, but seems to be not easy...

PR #180, opened 2018-08-23
Reads from config.groovy all docx/epub files

stoeps13:
Changed behavior, that not first docbook file is converted, instead script searches for epub and docx. These convert to docbook and then to the selected format.

### All Submissions:

Hi, don't know if these affects documentation. I use that script in one of my ci/cd scripts (which I will show on Froscon), so I want give you back the update.

### DIFF
--- a/scripts/pandoc.gradle
+++ b/scripts/pandoc.gradle
@@ -3,17 +3,20 @@ task convertToDocx (
         group: 'docToolchain',
         type: Exec
 ) {
-    // For now it's only taking the first input file that has docbook format specified
-    def sourceFile = sourceFiles.find { 'docbook' in it.formats }.file.replace('.adoc', '.xml')
-    def targetFile = sourceFile.replace('.xml', '.docx')
+    // All files with option `docx` in config.groovy is converted to docbook and then to docx.
+    def sourceFilesDocx = sourceFiles.findAll { 'docx' in it.formats }
+    sourceFilesDocx.each {
+        def sourceFile = it.file.replace('.adoc', '.xml')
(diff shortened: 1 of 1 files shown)


rdmueller:
perfect contribution! Thanx!

stoeps13:
Working on an error on my Linux box. The mkdirs() command is not working, so the files are not created. When I manually create docx and epub in build/docs it works like a charm.

stoeps13:
Workaround for Linux done, now the folders are created. Will create other PR. One question, how can I run generateDocbook automatically for all docx files in config.groovy?

The last 3 lines in pandoc.gradle are not enough.

rdmueller:
hm. The `generateDocBook` sources are here
[AsciiDocBasics.gradle#L164](https://github.com/docToolchain/docToolchain/blob/master/scripts/AsciiDocBasics.gradle#L164)

and they are structured the same way as `generateHTML` and `generatePDF` which work...

Have you added `"docbook"` to the `formats` sections of all your configured files in [Config.groovy](https://github.com/docToolchain/docToolchain/blob/master/Config.groovy)?

Does it only run over one file? Same problem with a `gradlew clean` before?

stoeps13:
Hi,
when I add docbook to the formats it works, but I search a way, that all files which have format docx selected in formats, are converted to docbook automatically without adding docbook to formats. 
Any chance to implement this? 

I thought this part would do that:
```
project.afterEvaluate {
    project.tasks.convertToDocx.dependsOn generateDocbook
    project.tasks.convertToEpub.dependsOn generateDocbook
```

But it's not important, I just add docbook and docx if I want to convert to docx (will not happen often).

PR #237, opened 2018-11-13
added hint to pandoc dep and formats in Config.groovy

ghp-dev:
### All Submissions:

* [X] Does your PR affect the documentation?
* [X] If yes, did you update the documentation or create an issue for updating it?

Once again, the current docs is technically correct: the necessary info (pandoc, docx format) is there, but very easy to overlook.

### Change to Documentation:


rdmueller:
cool. btw: have you seen the pandoc plugin?

it still needs some testing and the publishing of the final binary. but it will resolve the pandoc dependency automatically...

Issue #267, opened 2019-01-30
gradle build fails all tests on V1.0.0


SoftwareApe:
**Describe the bug**
gradle downloads all components but fails all tests.

**To Reproduce**
1. git clone --recurse-submodules <repo>
2. git checkout V1.0.0
3. cd docToolchain
4. gradle build

**Expected behavior**
Gradle builds all modules, executes tests and reports success.

**Screenshots**
```shell
gradle build

> Configure project :

docDir: .

mainConfigFile:
Config.groovy
inputPath: src/docs
inputFiles: [[file:test.adoc, formats:[html, pdf, docbook]], [file:manual.adoc, formats:[html, pdf]], [file:ppt/Demo.pptx.ad, formats:[html, revealjs]]] outputPath: build/docs
./build/docs/ppt/images

> Task :test

docToolchain.ExportChangeLogSpec > test creation of log file FAILED org.gradle.testkit.runner.UnexpectedBuildFailure at ExportChangeLogSpec.groovy:19
docToolchain.ExportMarkdownSpec > test conversion of sample markdown file FAILED org.gradle.testkit.runner.UnexpectedBuildFailure at ExportMarkdownSpec.groovy:20
docToolchain.GenerateDeckSpec > test correct generation of slide deck FAILED org.gradle.testkit.runner.UnexpectedBuildFailure at GenerateDeckSpec.groovy:17
docToolchain.GenerateDeckSpec > test skipped generation of slide deck FAILED org.gradle.testkit.runner.UnexpectedBuildFailure at GenerateDeckSpec.groovy:27
docToolchain.GenerateHTMLSpec > test correct handling of plantUML FAILED org.gradle.testkit.runner.UnexpectedBuildFailure at GenerateHTMLSpec.groovy:22
docToolchain.GeneratePdfSpec > test correct handling of plantUML FAILED org.gradle.testkit.runner.UnexpectedBuildFailure at GeneratePdfSpec.groovy:22
6 tests completed, 6 failed


FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':test'.
```

**Configuration**
 - OS: Windows 10, gradle via Cntlm proxy

**Additional context**

rdmueller:
can you please check the following:

go to the folder /docToolchain/resources

ist the reveal.js project available in the project? Has it been checked out?

SoftwareApe:
Hello Ralf,

[...]

PR #268, opened 2019-01-30
Add maven repository to sub builds so that tests can be executed

SoftwareApe:
Partially solves #267, now at least all files can be downloaded. Didn't use mavencenter() but maven with URL because it failed for some of the sub builds otherwise.

Now the tests execute, but some of them have errors. At least the gradle config seems to be more stable this way:

```shell
Condition not satisfied:

result.task(":generatePDF").outcome == SUCCESS
|      |                    |       |
|      |                    |       false
|      |                    UP_TO_DATE
|      :generatePDF=UP_TO_DATE
org.gradle.testkit.runner.internal.FeatureCheckBuildResult@245e3d61

	at docToolchain.GeneratePdfSpec.test correct handling of plantUML(GeneratePdfSpec.groovy:25)
```

And the problem that the sub builds don't take the proxy configuration from GRADLE_USER_HOME which the main build does also persists.

### DIFF
--- a/scripts/AsciiDocBasics.gradle
+++ b/scripts/AsciiDocBasics.gradle
@@ -4,8 +4,10 @@
 buildscript {
     repositories {
         jcenter()
+        maven {
+            url 'https://plugins.gradle.org/m2/'
+        }
     }
-
     dependencies {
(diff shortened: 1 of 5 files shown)


rdmueller:
thanx!

Issue #277, opened 2019-03-04
ConfluenceConfig - clarity and consistency


duschata:
Hallo All,
please have a look to some modifications I made here, I hope they are self-explaining

https://github.com/docToolchain/docToolchain/compare/master...duschata:duschata_issue-277

The var confluenceSpaceKey in asciidoc2Confluence.groovy is assigned, but never used in the context. I supposed that is not what you have meant. The suffix of the page and it's subpages would be a nice feature. A chapter like "myDocumentedCode-v3.2.1 1.2.3 MyPageTilte" is very hard to read.

According to this I've tried to achieve more clarity in the Config.groovy section docs. Starting from there some days ago, I was a little bit confused what exactly to do.

The
```
template_config/scripts/ConfluenceConfig.groovy
``` 
is probably outdated, so I added a little hint there

If you agree with these changes I would like to start a PR
  
Greetings from OF/Germany
Tom

rdmueller:
great! Thanx for the contribution!

Yes, please create a PR.

rdmueller:
PR is merged, so I guess this issue can now be closed...

PR #279, opened 2019-03-05
Duschata issue 277

duschata:
see: https://github.com/docToolchain/docToolchain/issues/277
Modifications are also documented in the doc section


rdmueller:
ok, I needed some time to check the change with the confluenceSpaceKey, but I guess you are right. This fixes a bug!
Thanx a lot!

You should also add yourself to the list of contributors!
https://github.com/docToolchain/docToolchain/blob/master/src/docs/manual/05_contributors.adoc

Issue #259, opened 2019-01-04
Initializing and building arc42EN template as described in the docs does not produce HTML output


pmatyszok:
**Describe the bug**
Initializing and building arc42EN template as described in the docs does not produce HTML output. Following steps described here: https://doctoolchain.github.io/docToolchain/#_how_to_install_doctoolchain in steps 2.1-2.3 ends with no output generated and message: 
```
:generateHTML NO-SOURCE

BUILD SUCCESSFUL in 1s
```

**To Reproduce**
1. ` git clone --recursive https://github.com/docToolchain/docToolchain.git docToolchain`
2.  `cd docToolchain/`
3. `rm -rf .git`
4. `rm -rf resources/asciidoctor-reveal.js/.git`
5. `rm -rf resources/reveal.js/.git`
6. `./gradlew -b init.gradle initArc42EN -PnewDocDir=arc42`
7. `./bin/doctoolchain arc42 generateHTML`
8. See:
```
:generateHTML NO-SOURCE

BUILD SUCCESSFUL in 1s

```
No build folder created.

**Expected behavior**
`build` folder created with sources generated inside.

**Configuration**
 - Windows 7 x64
- docToolchain run using "git-bash" - MINGW64

I am happy to help investigate this. Later I mangled a bit with settings, but it is hard to wrap the head arround the fact that they are splitted across `gradle.properties`, `Config.groovy` and can be overriden via commandline - not sure when and where the `inputPath` prop should be defined.

pmatyszok:
Yes, it did. The files were sucessfully unzipped:
![image](https://user-images.githubusercontent.com/3974142/50688413-acc5f680-1025-11e9-9c15-14e5ee9e01f7.png)

pmatyszok:
Missclick with closing the issue...

rdmueller:
thanx for reporting this bug.
can you show us your `Config.groovy`? It seems that the template to be rendered is not configured the right way.

pmatyszok:
I am attaching two files. One is `Config.groovy` from master directory, where docToolchain was check-out, and the other one, that is inside `arc42` directory. Btw. the `arc42` directory resides inside the docToolchain directory, but **not** in src directory (that is how the `initArc42EN` task created it).


docToolchain master config: [Config.groovy.txt](https://github.com/docToolchain/docToolchain/files/2727018/Config.groovy.txt)
[...]

PR #285, opened 2019-03-10
Fix for #259 and issues with standalone doctoolchain

jakubjab:
After this change, to generate docToolchain manual you need to specify inputPath and optionally pdfThemeDir:

./gradlew -PinputPath=./src/docs generateHTML
./gradlew -PinputPath=./src/docs -PpdfThemeDir=./src/docs/pdfTheme generatePDF

@rdmueller - please check if that doesn't break some of your assumptions/rules.


rdmueller:
Thanx for this contribution! More step further to make the interface more stable...

so, it does break something :-) I need some time to check if it is just the tests or more...
So please be patient...

jakubjab:
The tests were breaking for me even without this change, on plaim master. I should have fixed them before the change.

About the generation of the manual - maybe I would just document how to generate it using bin/doctoolchain (and copy to the right place) in the manual itself, and not strive to make this use case the simplest one.

rdmueller:
hm. the tests where working for travis. So it seems that you environment is different to the test environment.

I will check this.

Regarding the manual: there is this [copyDocs.sh](https://github.com/docToolchain/docToolchain/blob/master/copyDocs.sh) script which could be extended to generate & copy. I will take a look at this.

rdmueller:
I currently try to find out what is going on. The `exportExcel`- Task seems to export into the root folder.
I guess your tests where green because you already had those exported files in your `src/docs` folder and this only happens with a clean clone...

I now try to get some debug output but somehow it does not appear in the test report...

rdmueller:
So, the problem seems to lie in this line:
https://github.com/docToolchain/docToolchain/blob/master/scripts/AsciiDocBasics.gradle#L42

`srcDir  = "${docDir}/${inputPath}"`

you change sets `docDir` and ìnputPath` both to `.` but the tests expect that `docDir` is the path to the documentation project and `inputPath` is the path to the input files (`src/docs`)

I wonder how we can fix this. I will first try to fix the `exportExel` tests..

rdmueller:
Hi Jakub, sorry, it took me quite a long time to get back to this PR.

I've just upgraded the gradle version and fixed three tests (the ones which are failing with this PR).
Could you try to do a rebase on your branch so taht your change are on top of the master branch?

thanx, Ralf

jakubjab:
@rdmueller - after merge from master the tests are passing. However, I'm not sure what other things may fail. Can we have at least a list of manual tests to perform before merge? Then we could gradually make them automatic.

rdmueller:
hm. reagarding the manual tests - I would like to invest in some automated tests. So tell me about manual tests you do and I try to automate them.

but as for now, I guess you can easily test if your use cases still work and I will test mine.

If we are both happy, I am fine with merging this PR...

rdmueller:
... and if we find additional bugs, let's first write a test before fixing them...

rdmueller:
ok. my build breaks because of the input path.
old version: src/docs
this version: .

this results in `generateHTML` not finding the input files.

but I think I have a fix...

rdmueller:
argh. that broke lot of tests.
I think we need a better concept for  the configuration.

I think it would be great to have all configs in one file, independent of the gradle.properties file.
What do you think?

rdmueller:
next fix - now only three test break

jakubjab:
The latest version works for me, but I only use generateHTML and generatePDF. The tests are failing for me locally as well. I can try fixing them.

Regarding the configuration - I would be happiest if we could avoid configuration files at all, and pass all information in the command line. In gradle world this would be with -P I guess. Then it's flexible and different users can wrap it as they wish.

rdmueller:
regardings the config, let's start small.
I've created an issue for that: https://github.com/docToolchain/docToolchain/issues/310

jakubjab:
I have fixed the failing test cases. Seems that with the current version it's not possible to override the inputPath from the command line. But this could be the next step, once we agree on #310.

This version works for me as a standalone tool and fixes #259.
@rdmueller and everyone else interested - please check if your use cases are still working as expected with this change. If yes, then I think we could merge it.

rdmueller:
very cool, thanx! I will test it tomorrow...

rdmueller:
tests are green and changes look ok to me, so I will now sqash an merge.
Thanx for this fix!

PR #296, opened 2019-04-24
specify relative path to docToolchain

jagedn:
some users can be confused about where to locate `docToolchain` and what's the structure to follow

### All Submissions:

* [X ] Does your PR affect the documentation?
* [X ] If yes, did you update the documentation or create an issue for updating it? *Update*

### Your first submission

* [x] Have you added your name to the list of [contributors.adoc](https://github.com/docToolchain/docToolchain/blob/master/src/docs/manual/05_contributors.adoc)?

### Change to Documentation:


rdmueller:
interesting. please give me some time to think about this...

the `/bin`was omitted beause the docs expect that docToolchain is installed as cli and the `/bin` folder is on the path.

I don't use this approach. I always use docToolchain as gradle sub project. So I will try to see how this fits in the bigger picture.

jakubjab:
I think that is wrong, you should not need to specify `/bin` when using any tool. @jagedn - what is the root issue you see? Can you add `/bin` to your `PATH` instead? About some users confused about where to locate `docToolchain` - I think this should be addressed in the manual by clear separating 2 ways of using it - a) as gradle sub project, like @rdmueller does, b) as a standalone tool, as I'm using it. It could be even simpler if we could focus on a single goal of having a standalone tool, like `asciidoctor` itself.

rdmueller:
@jagedn : thanx for your patience with this PR.
I've just added you to the list of contributors. So this part ist solved :-)

regarding the relative path:
- when I use docToolchain in my projects, I use the `gradlew` commands
- other people like Jakub install doctoolchain in a separate folder and add the `/bin` to the path. That is what the docs intend to show.

I do understand that there is something missing to better understand this situation. I guess it would make sense to re-write the whole introduction to make things clearer.

What do you think?

rdmueller:
sorry for the wrong "wontfix" applied by the stale bot.
I currently try to work on my backlog. 
I do understand that the docs can be misunderstood.
So, I tried to apply a change to your propossed change which should clarify the situation and also prepare the docs for the next change - docToolchain can already be installed via sdkman as beta :-)

rdmueller:
let's close this as "stale". Please reopen when needed.

Issue #282, opened 2019-03-07
collectIncludes


rdmueller:
Collect includes should search for adoc files in the whole project, but it only searches in the task's project (`projectDir`) which is only the docToolchain folder if used as submodule.

The task should be extended to the whole project.

synesthesia:
Further, with docToolChain installed outside the project and added to the path, the `collectIncludes` task scans the docToolChain folder, _not_ the project folder

This is consistent I think with the use of `projectDir` as the source of files to scan at https://github.com/docToolchain/docToolchain/blob/master/scripts/collectIncludes.gradle#L15 

Perhaps this needs another property which defaults to the same value as `docDir` (for the common case where records such as ADR are in a subdirectory of the document directory), but which can be overridden on the command line with `-PscanDir=<other root>` ?

And then https://github.com/docToolchain/docToolchain/blob/master/scripts/collectIncludes.gradle#L15 might read
```
File sourceFolder = scanDir
```
?

what do you think @rdmueller ?

rdmueller:
thanx for your contribution!
As your PR changes the properties file, I need some more time to review it...
We jsut hat a change inthe propterties and I want to make sure that this doesn't interfere...

synesthesia:
Of course! To be honest as I get more into using docToolchain I am hacking more tasks to suit my use case, not just for paths but to adjust generated Asciidoc code. Feels like a codesmell so I think these should all be part of the bigger #279 and other similar “where do we go” conversations.

I think the conversation about subrepo v /bin is key

rdmueller:
that's great! most of the tasks ate hacked together to see if the use case fits and they are developed for just one use case.

so if you extend some of the tasks to fit a more general use case, it would be great if you could contribute them back.

rdmueller:
time flies by... I just managed to create another PR where I tried to address some of the issues...
Now I have to find some time to bring both together...

rdmueller:
Thanx for your contribution and thanx for the patience!
You PR has been merged!

PR #302, opened 2019-05-01
#282 possible fix collect includes

synesthesia:
Initial discussion around #282


scripts/AsciiDocBasics.gradle
```diff
@@ -26,7 +26,7 @@ if (docDir.startsWith('.')) {
     docDir = new File(projectDir, docDir).canonicalPath
 }
 if (pdfThemeDir.startsWith('.')) {
-    pdfThemeDir = new File(projectDir, pdfThemeDir).canonicalPath
+    pdfThemeDir = new File(docDir, pdfThemeDir).canonicalPath
```
rdmueller: this is likely to break things and does not have to do with the `collectIncludes` (which is ok for the moment)
But it seems to be the right thing...

scripts/collectIncludes.gradle
```diff
@@ -12,7 +12,26 @@ task collectIncludes(
     doLast {
         //let's search the whole project for files, not only the docs folder
         //could be a problem with node projects :-)
-        File sourceFolder = projectDir
+        if (docDir.startsWith('.')) {
```
rdmueller: hm. don't we already fix this in the `Asciidoctorbasics.gradle`-file? I have to check this...

scripts/collectIncludes.gradle
```diff
@@ -21,13 +40,19 @@ task collectIncludes(
                 if (!collections[type]) {
                     collections[type] = []
                 }
-                collections[type] << (file.canonicalPath - projectDir.canonicalPath)[1..-1]
+                println "file: " + file.canonicalPath
```
rdmueller: ok. this is only debug output, isn't it?
we should turn this into `logger.info`-statements, shouldn't we?

rdmueller:
I hope changing the branch works this way.
I will now resolve the conflicts and then merge to the new branch

rdmueller:
I've merged the two changes into an extra-branch to give it a try:
https://github.com/docToolchain/docToolchain/commits/collectIncludes

Unfortunately, it seems that through squashing etc git lost the info About your work. Hm. I guess I have to do it again...

Issue #194, opened 2018-09-24
integrate draw.io


rdmueller:

gernotstarke:
cool idea!

rdmueller:
just got some code up and running (not checked in yet) using the jgraphx / mxGraph library. 
It is kind of hard to work with it and the result is not usable (html in the shape text and most shapes do not render at all).
So I created a SO question with a cry for help :-)

https://stackoverflow.com/questions/53951770/render-png-image-from-draw-io-xml-file-in-java

rdmueller:
argh. just noticed that this was wasted time:

if you work with draw.io, just work with `.png` files:
- Export as PNG, tick the box 'include a copy of my diagram'
- when you change the diagram, just open this PNG
- all save operations will be performed on this PNG!

So, there is no need to do a conversion in order to integrate draw.io with asciidoc!!!! :-)

rdmueller:
see https://doctoolchain.github.io/docToolchain/#_exportdrawio

jaredmorgs:
I'm going to put in a bit more procedural direction in this section because I hit a wall when trying to follow along from what is currently written. I'll reference this bug when I'm writing it.

rdmueller:
@jaredmorgs thanx. I am curious about your thoughts

jaredmorgs:
@rdmueller I have submitted the PR. I have not yet built the docs just in case changes are required. 

I also don't have gradle configured on my PC, but I'm willing to do it if required. I just wanted to get the words written before worrying about building.

PR #332, opened 2019-06-27
Proposed an editing workflow for #194

jaredmorgs:
I have tried this process and it works well, but the original instruction set left a few too many questions.
I've proposed a workflow that folks can use to continue using draw.io.

I have not published the file just in case changes are required.

### All Submissions:

### Your first submission

* [ ] Have you added your name to the list of [contributors.adoc](https://github.com/docToolchain/docToolchain/blob/master/src/docs/manual/05_contributors.adoc)?

Not as yet. I will do that later on.

I have not published the file just in case changes are required.

I have added a note into #194 to re-build the docs once the change is improved.

### DIFF
--- a/src/docs/manual/03_task_exportDrawIo.adoc
+++ b/src/docs/manual/03_task_exportDrawIo.adoc
@@ -17,7 +17,7 @@ NOTE: If you are converting a Confluence page with embedded draw.io diagrams to 
 . Load the diagram you want to export from Confluence.
 . Click menu:File[Export as > PNG...]
 . In the Image modal, make sure that "Include a copy of my diagram" is selected.
-. Click Export to save the PNG file with the pattern `[file].dia.png`.
+. Click Export to save the PNG file with the pattern `[file].dio.png`.
 +
 NOTE: Specifying `.dia` in the name will help you identify PNG files containing embedded XML diagram source.
 +
--- a/src/docs/manual/03_task_exportDrawIo.adoc
+++ b/src/docs/manual/03_task_exportDrawIo.adoc
@@ -27,3 +27,5 @@ You now have a diagram that can be managed in source control, added to your docu
 
 NOTE: Draw.io offer free and open source desktop editors for all major operating system platforms. 
 See https://about.draw.io/integrations/ to find a desktop editor application compatible with your operating system.
+When you use the desktop version, just create your diagram with the `.png` or even better `.dio.png` extension and draw.io will always save your diagam as PNG with the source as meta data.
+
--- a/src/docs/manual/03_task_exportDrawIo.adoc
+++ b/src/docs/manual/03_task_exportDrawIo.adoc
@@ -19,7 +19,7 @@ NOTE: If you are converting a Confluence page with embedded draw.io diagrams to 
 . In the Image modal, make sure that "Include a copy of my diagram" is selected.
 . Click Export to save the PNG file with the pattern `[file].dio.png`.
 +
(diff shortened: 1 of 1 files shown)


src/docs/manual/03_task_exportDrawIo.adoc
```diff
@@ -1,15 +1,45 @@
...
+. In the Image modal, make sure that "Include a copy of my diagram" is selected.
+. Click Export to save the PNG file with the pattern `[file].dio.png`.
++
+NOTE: Specifying `.dio` (short for "**d**raw**io**") in the name will help you identify PNG files containing embedded XML diagram source.
++
+. Commit the exported PNG file to source control.
+
+You now have a diagram that can be managed in source control and edited using draw.io. 
+
+.Edit a diagram in Draw.io Offline editor mode
```
rdmueller: Do we need this section? I think it should be replaced with a hint to the Desktop version: https://about.draw.io/integrations/#integrations_offline

With this version, you can edit your diagrams offline and don't have to take care of duplicate files.

rdmueller:
Thanx! This is great! Better documentation is very valuable.

Two remarks: 
- I initially came up with the extension `.dio` for "drawio". As I do understand you used `.dia` for "diagram". I hope it is ok that I switched to `dio` and I added an explanation
- see my comment for the Desktop version. I think it solves the problems of the offline mode.

jaredmorgs:
Great pick up on the dio vs dia, and my apologies for confusing those two.
It's a really good pattern to use to identify these files.

Good tip about the desktop editor as well for draw.io. I'm going to give it
a try in the workflow where I work and see how it performs.

Did you want me to adjust the PR with the new instructions about the
desktop editor unless it's already been merged?

src/docs/manual/03_task_exportDrawIo.adoc
```diff
@@ -1,6 +1,6 @@
 :filename: manual/03_task_exportDrawIo.adoc
 ifndef::imagesdir[:imagesdir: ../images]
-:experimental:
```
jaredmorgs: :experimental: gives you access to the Menu and keyboard macros which I've used on line 18. Without that macro, the keyboard macro won't render correctly.

src/docs/manual/03_task_exportDrawIo.adoc
jaredmorgs: This is a much better idea. I just installed the Windows desktop version and its heaps better than the offline web editor. I was just going with what I thought was suggested in the original text.

src/docs/manual/03_task_exportDrawIo.adoc
rdmueller: ah. ok. didn't see that. let's put it back in...

rdmueller: ah - I've seen you've already put it back in. perfect

rdmueller:
> Did you want me to adjust the PR with the new instructions about the
> desktop editor unless it's already been merged?

I think best would be if you first check out the Desktop version of draw.io. And then let's see if we still need the section about offline mode (IMHO, the Desktop version is better to use).

jaredmorgs:
Yep already done it and I agree the Desktop version is superior. I adjusted
the procedure accordingly.

rdmueller:
perfect! Welcome to the ever growing lsit of contributors!
Feel free to add yourself to the list of contributors within the docs through another PR.

PR #340, opened 2019-08-14
Collect includes2nd try

rdmueller:
see #282


rdmueller:
problem seems to be caused by oracleJDK8
Tests pass for openJDK. So let's contitue....

Issue #342, opened 2019-08-19
Proposal: Convert keywords to Confluence labels


dierk42:
When publishing to Confluence I'd like to attach labes (=keywords) to pages. This helps creating reporting pages and improves searching results.

In an asciidoc header I define :keywords: which become a meta tag 'keywords' in HTML.

In my version of asciidoc2confluence.groovy I split the keywords into a list and add the labels after(!) the creation of a page. Confluence does not allow to include labels at the time of creating a page.

My quirking the script works for me. But as I'm not an experienced Groovy programmer -- have look at it if you think this feature useful.

[asciidoc2confluence.txt](https://github.com/docToolchain/docToolchain/files/3515982/asciidoc2confluence.txt)

rdmueller:
^hey, this sound like a good idea!
Do you mind to send you code as a Pull-Request? This would make it easier to review it as diff...

sabatmonk:
this is in master, should it be closed?

rdmueller:
thanx. yes, I guess it can be closed.

PR #343, opened 2019-08-19
Create asciidoc2confluence.groovy

dierk42:
Proposal #342: Add keywords as Confluence labels. All generated pages get the same list of labels attched.

Good for reporting and searchin in Confluence

### All Submissions:

* [x ] Does your PR affect the documentation?


dierk42:
My first pull request on GitHub. Sorry if somethings are wrong.

rdmueller:
happy to have this PR! Thanx!

rdmueller:
something is strange about this PR...
ah! You didn't overwrite the original script. This is why I can't see a diff.
Could you please move the `asciidoc2confluence.groovy` script to the `scripts/.` folder in your repository and do another commit/push to the branch?
This will automatically update the PR and give us the right diff...

dierk42:
My fault, sorry. Should be correct now.

scripts/asciidoc2confluence.groovy
```diff
@@ -386,20 +438,26 @@ def pushToConfluence = { pageTitle, pageBody, parentId, anchors, pageAnchors ->
 
     def pages
     trythis {
-        def cql = "space='${confluenceSpaceKey}' AND type=page AND title~'" + realTitle(pageTitle) + "'"
+        // #342-dierk42: Colons in title irritate Confluence's lucene search engine
+        // (not really part of #342 but useful)
+        def sPageTitle = pageTitle.replace(':', '')
```
rdmueller: hm. shouldn't we escape the ':' as special character? And aren't there more such special characters?

dierk42: Agreed. The colon was a quick hack for me. As we are searching for titles all characters with a special meaning should be escaped/deleted/masked. This is valid for colon, the relational operators like '<', '!=' and so on. An of course the Wildcard characters '*' and '?'.

As your using the CONTAINS-operator in the cql (title~...) it should be sufficient to delete all these characters. The rest are words or numbers that are ok. 

I have no experience what happens when you use the '=' for the title search. Perhaps then you'ld need the special characters? 

To cope with these special characters would make the routine more robust I think.

dierk42: The other question regarding adding documentation: Of course I can do it. Where in the docs would you like to have it?
Some documentation is needed because Confluence allows no spaces in the keywords. One schould know this fact. But something lile "this_is_a_label" would be correct.

We could improve the part where the list of keywords in asciidoc2confluence.groovy is built up and replace all whitespace by '_', after the trim operation. (line 612 ff in my version)

rdmueller: regarding the special characters: I just did a quick google search and there seem to be some open bugs with this: https://jira.atlassian.com/browse/CONFSERVER-55559
...and I remember that we already have some fixes in place like chacking for the number of search results.
So let's leave it as it is for the moment...

rdmueller: regarding the docs: take a look at https://doctoolchain.github.io/docToolchain/#_publishtoconfluence
there is an orange button in the upper right corner which will take you to the sources...
Any documentation is appreaciated!

rdmueller:
thanx!!!

PR #344, opened 2019-08-20
Improving doctoolchain.bat. Introduced call gradlew and switch back t…

dierk42:
On Windows: Using doctoolchain.bat like "doctoolchain . generateHTML" from a document source path other than the default (e.g. c:\mysources\myproject\docs)  does not swith back to the document source path.

Reason: 

In doctoolchain.bat the control goes to gradlew-script. When it ends the working path switched to the installation of doctoolchain and not back where I started.

I added a 'call' for gradlew and at the end of doctoolchain.bat reassure that the document working dir is activated so you stay where you came from.


rdmueller:
that's great. Thanx!

Issue #348, opened 2019-08-23
publishToConfluence not working in standalone Confluence server


emfanitek:
**Describe the bug**
The Confluence Server cql seems not to be interpreted the same as the cloud-based confluence.
In the `pushToConfluence` method, the check against the title uses the `~` (CONTAINS) operator.

When the title has a number in it, the server does not match the title.

Example:

this will match

> cql=space=TECHDOC AND type=page AND title = "GW - 2. Deployment View"

this won't match  
> cql=space=TECHDOC AND type=page AND title ~ "GW - 2. Deployment View"

When the title has no number, the contains operation matches.

There is no documented specificity to numbers in text searches in the confluence documentation:

https://developer.atlassian.com/server/confluence/performing-text-searches-using-cql/

**To Reproduce**
1. Use standalone confluence server
2. Publish to confluence a page with sections containing numbering
4. Script cannot find existing pages and will complain later when performing the create, that pages exist

**Expected behavior**
Existing pages are updated.

**Screenshots**

**Configuration**
 - OS: linux
 - Confluence version: 6.15.7

**Additional context**
I wonder if using the CONTAINS operator is really necessary. My query worked OK when I replaced CONTAINS with EQUALS.

rdmueller:
thanx for this issue.
The search through the API is a bit flaky. We already switched from the equals operator to the contains operator (and then check the result set to match the exact result if we got more than one).
The equals operator cause problems in the past with some special characters. I don't remember any more which ones but I guess it was the `:` or `.`.

So, it seems to me that changing the search string is not the right solution because to use the API is not the right solution in the first place. I even found bug reports which are years old.

I think the right solution would be to first fetch the root document and then fetch all child elements in a recursive way to create an in memory model. This would allow to do an exact search in memory.

It would also allow us to implement additional features like a case insensitive seach. (You often correct the case of a letter and this now results in a new page because the old will not be found anymore)

what do you think?

LuisMuniz:
Hi Ralf, 

An atlassian consultant was recommending to use the GET api simply:

https://<server>/rest/api/content?spaceKey=<Sspace>&type=page&title=GW%20-%202.%20Deployment%20View2

But then we might get into trouble with special characters too.

Downloading the whole space pages does not seem to be too expensive i guess.

I can upgrade my pull request o implement that, if you wish, and use case-insensitve matching.

rdmueller:
thanx for this wonderful PR.
[...]

PR #349, opened 2019-08-25
Replace Contains GQL operator with simple EQUALS operator to mathc pa…

LuisMuniz:
Allow publishing to a standalone confluence server, by changing the operator used to find an existing page matching the realTitle from contains (`~`) to EQUALS (`=`)

I did not create new tests for my submission (publishing to confluence is a hard thing to test).

however, my change is backwards compatible as I tested it both on cloud and standalone confluence.

### DIFF
--- a/scripts/asciidoc2confluence.groovy
+++ b/scripts/asciidoc2confluence.groovy
@@ -23,7 +23,7 @@
  */
 
 /*
-    Additions for issue #342 marked as #342-dierk42 
+    Additions for issue #342 marked as #342-dierk42
     ;-)
 */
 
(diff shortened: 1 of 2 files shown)


LuisMuniz:
related to issue #348

rdmueller:
please see my comment in #348 .
I would accept the PR immediatly if this would be implemented as option with the old behaviour as default. Other wise, I currently don't know what it might break.

Maybe we should add some tests. We could put a confluence instance in a container or even better, we do have a free cloud instance to which I can give you access (just write me a mail: ralf.d.mueller@docs-as-co.de)

LuisMuniz:
Hi,

I'm not sure if you can create a development confluence server, that would not need a license. I used the atlassian docker image, and the default takes you through a wizard that asks for a key.

Not ideal for CI.

I will cancel this pull request and create a new one, that will pull all the page matadata to do the matching in-memory.

LuisMuniz:
Hi, weren't you going to refuse this PR and instead merge #352 ?

PR #359, opened 2019-09-23
Paginate the result returned by confluence API as many times as neces…

### LINKED ISSUE #358: publishToConfluence fails to update existing pages when size of space is bigger than 25 pages
**Describe the bug**
When updating existing pages in confluence, the retrieveAllPages method fails to paginate the full resultset. Because by default the size of the result set is 25, when the number of total pages that a confluence space has, exceeds that number, the retrieveAllPages method does not initially return it. 

Subsequently, the logic assumes this is a new page, and ultimately fails with a code 400 when the script tries to CREATE a new page, instead of updating an existing page.

**To Reproduce**
0. Have a confluence space bigger than 25 pages.
1. publishToConfluence
2. publishToConfluence
4. See error "A page with this title already exists"

**Expected behavior**
The pages should be properly updated

**Screenshots**

**Configuration**
 - Confluence Server 6.15.9

**Additional context**

LuisMuniz:
fixes #358

### DIFF
--- a/scripts/asciidoc2confluence.groovy
+++ b/scripts/asciidoc2confluence.groovy
@@ -231,31 +231,47 @@ def retrieveAllPages = { RESTClient api, Map headers, String spaceKey ->
         allPages
     } else {
 
+        boolean morePages=true
+        int start=0
         def request = [
                 'type'    : 'page',
                 'spaceKey': spaceKey,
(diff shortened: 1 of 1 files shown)


rdmueller:
Thanx! 👍

PR #362, opened 2019-10-04
calculate localHash after prefixing extraPageContent

TimRiemer:
The MD5 for the localPage was calculated before the extraPageContent was prefixed to the localPage. Therefore the hashes didn't differ, if there weren't any other page changes in the local HTML.

With this change it is now possible to just change the extraPageContent banner and push it to Confluence.


rdmueller:
very cool. thanx!

PR #379, opened 2019-10-27
Update remaining dependency versions except the asciidoctor related ones

dbast:
Doing this while getting more familiar with the toolchain.

Updating asciidoctor dependencies to versions that play well together while not breaking anything seems to be not trivial or even requires some task code changes due to API updates in:
* asciidoctor-gradle-plugin
* asciidoctor.convert
* asiidoctorj-pdf
* asciidoctorj-diagram

Seems to be an interesting task to explore to see how everything plays together. But that would anyway go via different PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: groovy
 jdk:
   - openjdk8
   - openjdk11
+  - openjdk12
 
 addons:
   apt:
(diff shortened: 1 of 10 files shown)


rdmueller:
Thanx!
One feature of docToolchain is that it bundles versions which do not break any features. That is why I only upgrade version when I need a feature of a newer version in one of the libraries.

But as the checks are all green and I see from your comments ("updated to latest non-breaking vers.") that you did the upgrade carfully, I am happy to merge this request.

Many thanx!

rdmueller:
interesting - it now breaks the jdk11 build....

rdmueller:
local test with jdk11 works... let's define that travis has a problem ;-)

Issue #375, opened 2019-10-25
Publication to Confluence fails with MissingMethodException


dArignac:
**Describe the bug**
I'm trying to publish to Confluence, but it fails with `No signature of method: asciidoc2confluence$_run_closure13.call() is applicable for argument types`.

**To Reproduce**
1. checkout doctoolchain master at the current latest commit da364e921
2. use an arc42 project
3. publish to confluence `doctoolchain.bat . publishToConfluence` 

**Expected behavior**
Publication is successful.

**Stacktrace**
```
λ doctoolchain.bat . publishToConfluence --stacktrace

> Configure project :
index.adoc

> Task :publishToConfluence FAILED
publish C:\Users\alex\Projects\architecture\./build/html5/index.html
1. Introduction and Goals

FAILURE: Build failed with an exception.

* Where:
Script 'C:\tools\docToolchain\scripts\publishToConfluence.gradle' line: 29

* What went wrong:
Execution failed for task ':publishToConfluence'.
> No signature of method: asciidoc2confluence$_run_closure13.call() is applicable for argument types: (String, org.jsoup.select.Elements, Integer, LinkedHashMap, LinkedHashMap...) values: [1. Introduction and Goals, [<div class="sectionbody">
  <div class="admonitionblock important">
  <table>
  <tr>
  <td class="icon">
  <i class="fa icon-important" title="Important"></i>
  </td>
  <td class="content">
  TODO
  </td>
  </tr>
  </table>
  </div>
  <div style="page-break-after: always;"></div>
  </div>], ...]
  Possible solutions: any(), any(), grep(), collect(), dump(), find()

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':publishToConfluence'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter$3.accept(ExecuteActionsTaskExecuter.java:166)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter$3.accept(ExecuteActionsTaskExecuter.java:163)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:191)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:156)
[...]

PR #381, opened 2019-11-01
Clarified usage and type of ancestorId #375

dArignac:
Source #375 
This fixed the documentation to clarify that the `ancestorId` shall be a of type `String`. 

I did not build the docs for several reasons:
* the `feedback.adoc` cannot be included (on every page), didn't find out why (it was already an issue when I generated docs based on master, also in the published version) - and that would leave all generated HTML with my local path I don't want in the repo
* the doc generation changed a lot of HTML properties - not sure why (I'm on Windows btw.), e.g. 
```
-<img src="images/demoPlantUML.png" alt="demoPlantUML" width="246" height="236">
+<img src="images/demoPlantUML.png" alt="demoPlantUML" width="231" height="236">
```

### DIFF
--- a/Config.groovy
+++ b/Config.groovy
@@ -56,7 +56,7 @@ confluence = [:]
 // Attributes
 // - 'file': absolute or relative path to the asciidoc generated html file to be exported
 // - 'url': absolute URL to an asciidoc generated html file to be exported
-// - 'ancestorId' (optional): the id of the parent page in Confluence; leave this empty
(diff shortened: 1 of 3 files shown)


rdmueller:
thank you very much! I will regenerate the docs in a moment...

rdmueller:
hm. it also took me a while to generate the docs. another issue to solve...

but here it is with your name in it :-)
https://doctoolchain.github.io/docToolchain/#_publishtoconfluence

PR #385, opened 2019-11-19
sort filenames to avoid jumping chapters on some environments

stehlih:
### All Submissions:

In our Linux & Docker environment the list of files was not sorted by name. So we get "jumping chapters" in the documentation. To avoid this problem the file list is now sorted by filename.
On Windows environments we can't reproduce this behavior.

* [N] Does your PR affect the documentation?
* [-] If yes, did you update the documentation or create an issue for updating it?

### Changes to Core Features:

* [N] Have you written new tests for your core changes, as applicable?
* [Y] Have you successfully ran tests with your changes locally?

### Change to Documentation:

* [N] Did you build the docs and copy them to the `/docs` folder?
* [N] If not, did you create an issue for doing so?


rdmueller:
thanx. That is a great change to make it more stable!

Issue #315, opened 2019-05-22
Confluence page does not render code callouts properly


LuisMuniz:
One of my favorite features of asciidoctor is to be able to add callouts to code blocks, like this:

    [source,yaml]
    .Add lastpass-cli to your docker-compose.yml
    ----
      cli:
        image: cli:v2.0.5
        entrypoint: "tail" <1>
        command: ["-f","/dev/null"] <1>
    ----
    <1> This will override the default entrypoint `cli.sh` and just hang with a `tail -f /dev/null` to prevent docker from exiting. See link:../infrastructure/cli-docker.html[Article on cli] for details.

However Confluence does not seem to support this in their own code blocks.

When asciidoc2confluence rewrites the generated HTML to translate them to Confluence code blocks, the resulting code blocks show as inlined HTML in the code, something like this:

    cli:
      image: cli:v2.0.5
      entrypoint: "tail" <i class="conum" data-value="1"></i><b>(1)</b>
      command: ["-f","/dev/null"] <i class="conum" data-value="1"></i><b>(1)</b>

Is it possible to just disable rewriting the code block, so that it is not recognized by confluence as a code block, and thus not suffering from their limitations?

**To Reproduce**

1. Add a code block with callouts to your asciidoctor file
2. publishToConluence
4. See resulting Confluence page

**Expected behavior**
That the uploaded confluence page shows callouts like a normal asciidoctor rendered HTML page.

rdmueller:
good idea, we can give it a try. But I guess that the code will then not get syntax highlighted...

you can give it a try by commenting out this line https://github.com/docToolchain/docToolchain/blob/master/scripts/asciidoc2confluence.groovy#L336

It should do the trick that the code block does not get rewritten...

If you do so, could you post screenshots of both rendereings?

LuisMuniz:
Sure I'll post the screenshots

LuisMuniz:
This is the rendering produced by Asciidoctor:

![asciidoctor-rendering](https://user-images.githubusercontent.com/8201284/61383005-22046500-a8ae-11e9-8343-cd3f5f7f6090.png)

This is the rendering in Confluence with the current script:
![default-code-conversion](https://user-images.githubusercontent.com/8201284/61383053-38aabc00-a8ae-11e9-9c2b-e9f467ca6fb0.png)

This is the rendering in Confluence when shorting out the `rewriteCodeblocks` method:
[...]

PR #397, opened 2020-01-22
Handle callouts on publishToConfluence

bohni:
rewriteCodeBlock rewrites callouts as comments in the code.

Issue: #315


rdmueller:
thanx. I have no time to check if it works, but if you say it works for you... let's see what others say...

Issue #428, opened 2020-06-04
Missing  authentication credentials when publishing to confluence - http error 401


PeterStange:
Issue happens whenever the publishToConfluence task is used to upload HTML pages to a confluence area with user access restrictions. In case an API key is required, the connection is refused.
Even the API key is set in configuration, the header of the upload request does not contain the key/value of it.
So, it is rejected because of missing authentication.
`What went wrong:
Execution failed for task :docToolchain:publishToConfluence. 
java.lang.Exception: missing authentication credentials`

After code review, the reason is detected:
Can be fixed by adding the missing header field to all REST API requests.
Proper pull request is already prepared.

 - OS: Windows 10, Ubuntu 16.04.5 LTS

rdmueller:
interesting. Where is this header field documented?
I always use the basic auth with API-token as described here: https://developer.atlassian.com/cloud/confluence/basic-auth-for-rest-apis/
This way, you use your user-id + API token instead of user-id + password.

dr460neye:
Actually i use it since weeks without any issues. 
It works for on-premise as also on the atlassian cloud.

The major difference is, that for on-premise i use:
User + Password

For Cloud i use:
Emailaddress + API Token

PeterStange:
> 
> 
> interesting. Where is this header field documented?
> I always use the basic auth with API-token as described here: https://developer.atlassian.com/cloud/confluence/basic-auth-for-rest-apis/
> This way, you use your user-id + API token instead of user-id + password.

We also use a "Atlassian Confluence" based Wiki software. But, for us the Basic Auth header  identifies the user and allows the fine-grained permission handling on user level, and therefore a user/password credential is required. In addition to this, the  API Gateway need to pass an API key. The key is sent in the HTTP header. 
The way to user the user-id+API token in the Basic Auth header is not working.

rdmueller:
interesting - I didn't know about this configuration.
So thanx for the PR, but I guess we should clarify in the comments how and when to use it...

PeterStange:
Yes, agree. It need to be documented. I'll add a proper description.

PeterStange:
After rough documentation in the manual is taken over #437 into master, this issue can be closed.

PR #429, opened 2020-06-04
Fix for #428 publishToConfluence authentication failure

PeterStange:
As reported in issue  #428, the configuration of publishToConfluence is not used for the APIkey representation in the HTTP header. So, requests were rejected in case it is required with reason: missing authentication credentials.
This pull request fixes the issue by adding the apikey key/value pair if the config parameter: 'config.confluence.apikey' is set, otherwise no additional header field is added and previous behaviour is retained.

### All Submissions:

Documentation not affected.


rdmueller:
thanx!

Issue #439, opened 2020-06-18
exportEA do not use docDir for export folder


PeterStange:
**Describe the bug**
A doc-as-code project is set up using docToolchain as an external repository. 
[docToolchain]/bin folder is added to PATH environment variable.
Running on windows.
From inside the docDir directory the command is called:"doctoolchain . exportEA"
The export is executed.
expected:
Instead of expected location "[docDir]/src/docs/images/ea",
diagrams are exported into "[docToolchain]//src/docs/images/ea".

reason:
exportEAP.vbs does not use the docDir as a base.
exportEAP.vbs uses the current directory as base.
The current directory is set to the script dir instead.

Is there any workaround know to avoid this? Maybe a missing configuration?

rdmueller:
Thanx for the PR!!! I guess this issue can now be closed?

PeterStange:
Thanx for clarification of the test failure. Yes, we can close it.

PR #441, opened 2020-06-29
Add parameter exportPath and searchPath to exportEA module

PeterStange:
As described in issue #439, the export does not work as expected in case the docToolchain repository is not used as a submodule, but used as a standalone tool. In this case the configured docDir wasn't used as base directory.
This is fixed by this pull request. 
In addition the path to export the diagrams can be set as an optional parameter. Same for the path where to look for eap or eapx project files. Both paths are relative to the docDir path.
The sample Config.groovy and documentation is updated accordingly.
A method is added tp clean the filenames in case it contains invalid characters.

### DIFF
--- a/Config.groovy
+++ b/Config.groovy
@@ -143,5 +143,10 @@ exportEA.with {
 //                    "{A237ECDE-5419-4d47-AECC-B836999E7AE0}",
 //                    "{B73FA2FB-267D-4bcd-3D37-5014AD8806D6}"
 //                  ]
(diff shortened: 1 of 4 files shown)


rdmueller:
thanx for this valuable PR. I'll check what's wrong with the tests....

rdmueller:
seems to be a network problem...

rdmueller:
Great code! I never took the time to check how arguments could be handed over to the VBS... :-|

Issue #454, opened 2020-08-05
Ability to save Jira request(s) results to the Excel file(s)


ewoks:
**Is your feature request related to a problem? Please describe.**
It is nice that we are able to save Jira query results in Asciidoc format, but a lot management people prefer Excel for different tabular data, especially if it is about Jira reports and similar "current status" data. 

**Describe the solution you'd like**
It would be great to use data pulled from Jira and save these in the Excel file with a given name.

Two additional flags would be added to the JIRA configuration:

- **saveAsciidoc**={true,false} _// when true results of the request to Jira API would be saved in a {filename}.adoc file_
- **saveExcel**={true,false} _// when true results of the same request to Jira API would be saved in a {filename}.xlsx file_

In addition to ability to save to Excel files, it would be good to provide a way to make multiple queries to JIRA and save results in respective files

**Additional context**
This should be completed after [Jira configuration refactoring](https://github.com/docToolchain/docToolchain/issues/451) ticket is closed (✔)

ewoks:
WIP

ewoks:
Can be closed with after #458 merge

ewoks:
Can be closed with as #458 is merged

PR #458, opened 2020-08-07
Feature/export jira to excel

ewoks:
PR regarding https://github.com/docToolchain/docToolchain/issues/454

- Support for multiple Jira queries
- Basic support for custom fields in Jira results
- Saving to Excel files

### All Submissions:

The source of the documentation can be found in `/src/docs/manual`.
To "publish" it, execute `./gradlew exportContributors && ./gradlew && ./copyDocs.sh`.
This will convert the `.adoc` file to HTML and copy them to the right folder so that github pages will pick them up.

### Change to Documentation:

* [x] Did you build the docs and copy them to the `/docs` folder?
* [x] If not, did you create an issue for doing so?

### DIFF
--- a/Config.groovy
+++ b/Config.groovy
@@ -189,10 +189,36 @@ jira.with {
     // the label to restrict search to
     label = 
 
-    // Jira query
-    jql = "project='%jiraProject%' AND labels='%jiraLabel%' ORDER BY priority DESC, duedate ASC"
+    // Base filename in which Jira query results should be stored
+    resultsFilename = 'JiraTicketsContent'
+
+    saveAsciidoc = true // if true, asciidoc file will be created with *.adoc extension
+    saveExcel = true // if true, Excel file will be created with *.xlsx extension
+
+    /* 
+    List of requests to Jira API:
+    These are basically JQL expressions bundled with a filename in which results will be saved.
+    User can configure custom fields IDs and name those for column header,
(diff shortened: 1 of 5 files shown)


ewoks:
Added ticket for documentation update

ewoks:
@rdmueller do you have some suggestions on this?

rdmueller:
the feature sounds great.
Do I understand the code in the right way that it is a breaking change? So if users used the old feature, they will have to update their config?

ewoks:
That's correct. Old configuration is incompatible. Can you suggest how we cause the least pain for the users?
Migration is pretty simple, as I added comments at the bottom of the Config.groovy but not sure about the best way to call users attention to it. Documentation with snippet of new configuration?

rdmueller:
the code could check if the `requests`-map is empty and if so, it could populate it from the old `jql`-config. What do you think?

ewoks:
Sounds like a good transition solution. I will try to work on it in the following days and ping you when it is ready. 
I would still emphasize that this is legacy configuration that would be removed in the future. Is that ok?

rdmueller:
> I would still emphasize that this is legacy configuration that would be removed in the future. Is that ok?

yes, let's mark it as depracated and output a warning. But I think we should support it for 6-12 month

rdmueller:
looks good to me. thanx for your work!

PR #473, opened 2020-09-07
Feature/export jira sprint changelog

ewoks:
This PR is "extension" of work done to export data from Jira with a specific focus on generating "change logs" / **"release notes" for each sprint based on Jira issues** assigned to that specific sprint.

Issues will be listed in a resulting excel/asciidoc file(s) as simple as `Key<url>`, `Summary` but users can configure if they prefer to list ticket status, type and assigned user (typically one who resolves the issue).
As in different organisation ticket status can have different conventions, user is flexible to define statuses of ticket for a changelog, as well as states in which sprint need to be for its changelog based on tickets to be generated.

p.s. I'm aware of simple change log generation based on `git log`.  In addition to that task, I assume and hope some other teams/organisation prefer to generate change log based on Jira sprints, so I wanted to offer this possibility. IMHO both methods have advantages and disadvantages.
In the future, both these methods could be used to cross-reference data from both sources and generate even more polished changelogs.

### All Submissions:

### New Feature Submissions:

2. [x] Does your submission pass all tests? (see travis check)

### Change to Documentation:


rdmueller:
thanx! I guess I have to update the block diagram soon!

PR #475, opened 2020-09-10
Feature/folder per jira task

ewoks:
This PR adds convenience configuration to separate results of `exportJiraSprintChangelog` & `exportJiraIssues` tasks, which can be useful in cases there are several result files

In addition PR updates project changelog


rdmueller:
thanx!

PR #476, opened 2020-09-11
Feature/credentials params

ewoks:
This PR takes care of **secrets in configuration files** and offers users clean possibility to pass credentials as parameters to specific docToolchain commands.

For example, to pass Confluence credentials use:
`./doctoolchain ../ publishToConfluence -PconfluenceUser=admin -PconfluencePass=secretPassword42` 

or, to pass Jira credentials:
`./doctoolchain ../ exportJiraIssues -PjiraUser=admin -PjiraPass=jiraPassword`

As often in production environment credentials for Jira & Confluence are the same, users could take advantage of the highest priority `username` & `password` parameters, which can be used for both endpoints.

After this PR, docToolchain can be safely used in CI/CD with for example Jenkins secrets without risk that credentials will be exposed in project/documentation/cofinguration repository

Parsing Config.groovy now indeed happens just once, so we could safely remove copy pasted code which parsed configuration in few different files. 

 
p.s. Documentation needs to be updated ASAP with separate ticket #477

### DIFF
--- a/Config.groovy
+++ b/Config.groovy
@@ -98,13 +98,16 @@ confluence.with {
 
     pageSuffix = ''
 
-    // username:password of an account which has the right permissions to create and edit
-    // confluence pages in the given space.
-    // if you want to store it securely, fetch it from some external storage or leave it empty to fallback to gradle variables
-    // set through gradle properties files or environment variables. The fallback uses the 'confluenceUser' and 'confluencePassword' keys.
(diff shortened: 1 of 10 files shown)


ewoks:
Tested with local instances as well as on CI/CD with docker image, but lets wait for green travis tests just in case

rdmueller:
thanx for this PR!

Issue #487, opened 2020-11-04
Uploading to confluence with configured proxy fails on attachment upload


rdnp:
**Describe the bug**
When executing the asciidoc2confluence script, on updating images, the script fails when a proxy is configured.
The error message shown is InvocationTargetException with unknown host.
It seems the proxy is not used for the update operation of that single image.

**To Reproduce**
1. Configure a proxy for confluence in the Config.groovy
2. Execute asciidoc2confluence.groovy
3. Change the content of an image to be updated

**Expected behavior**
The image should be updated

**Screenshots**
N/A as it is a command line invocation

**Configuration**
 - OS: Windows
 - Browser N/A 
 - Relevant SDK Versions JDK 1.8, Groovy 2.4.3 run through Maven build

PR #488, opened 2020-11-04
Fix issue with proxy while updating images

rdnp:
### All Submissions:

* [ ] Does your PR affect the documentation? --> No
* [ ] If yes, did you update the documentation or create an issue for updating it? --> N/A

This PR suggests a fix to issue #487. The fix works in our environment, where a proxy is being used.

With this PR I intend to create awareness for the issue and suggest a possible solution. However, I did not have time to add automated regression tests or do a thorough analysis of the code base. So it might be needed that someone additionally checks for any side effects.


rdmueller:
thanx for your contribution. the code looks good to me, so I don't think we need tests for it.

rdmueller:
oh, I thought I merged this already two days ago. sorry for the delay.

Issue #490, opened 2020-11-06
Export the Enterprise Architect built-in Project Glossary to be used as glossary inside asciidoc 


PeterStange:
There is a built-in Project Glossary available in Enterprise Architect. Generating documentation out of EA can directly add it to the documents. Same glossary is expected to be used for docToolchain based documentation.
Currently, diagrams and element notes were exported only.

**Describe the solution you'd like**
A configurable option to export the project glossary as part of the exportEA script.
The option shall allow to define the output format (in asciidoc notation) and to filter the exported glossary terms by type.
The output format specifier shall support all supported elements of a project glossary, but the use of each single element shall be configurable.
AS usual glossary terms shall be sorted in alphabetical order.

The existing vbs script is used to export the glossary terms in a JSON formatted syntax.
As sorting and file handling is well done and easy to use in groovy, this parts shall be done in the gradle file of exportEA.

In case multiple EA projects are involved, one single glossary file per project plus one global glossary file over all single project glossaries shall be created. All exports are filtered and sorted in the same manner.

The documentation needs to be updated accordingly.

PeterStange:
#491 fix #490

PeterStange:
Thanks, for the fast pull request acceptance. So it makes fun.

PR #491, opened 2020-11-06
Add new exportEA feature to export project glossary filtered & sorted

PeterStange:
This pull request refers to issue #490

It is a new feature of the exportEA script.
It enables the exportEA script to run an export of the EnterpriseArchitect project glossary.
The output format can be configured to be according to asciidoc notation.
The types of elements used to generate the exported asciidoc formatted is configurable.
A new glossary subfolder will be added to the exportPath.
Multiple glossary files will be generated in case of multiple project files are found.
One global glossary file is generated by merging all single glossary files.
The generated  files can be used as part of the resulting documentation depending in individual format of the user.

Documentation is added and reformatted manual of exportEA.

### DIFF
--- a/Config.groovy
+++ b/Config.groovy
@@ -135,11 +135,19 @@ confluence.with {
 //Configuration for the export script 'exportEA.vbs'.
 // The following parameters can be used to change the default behaviour of 'exportEA'.
 // All parameter are optionally.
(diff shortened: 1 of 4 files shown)


rdmueller:
thanx! This looks like a nice feature! Great that it already comes with docs :-)

Issue #450, opened 2020-07-15
ExportEA: Diagram Notes are not exported


AndreaMcls:
The Diagram Objects notes are exported as expected, however the Diagram Notes are not exported.

rdmueller:
I see, I guess the docs are missing an important detail:
you have to prepend `{adoc:filename}` to your notes in order to export them to `filename.adoc`. All notes with the same `filename` will be appended.

Does this info help?

AndreaMcls:
Yes, it is correct that only elements notes with the prefix `{adoc:filename} `are exported.  It is also correct that the diagram elements are exported in a .ad file and  all the elements belonging to a diagram are appended in one .ad file. This works for the elements of a diagram but the note of the diagram itself is not exported.
The fix of this is just one line of code in `exportEAP.vbs`, I will do a pull request as soon as I can.

rdmueller:
ah! cool. I never had this use case. Thnax for the PR!

AndreaMcls:
Before proceeding with the PR I need approval from my employer, I'm not sure how long it will take.

AndreaMcls:
thank you @PeterStange for the PR

PR #501, opened 2020-11-30
Add the export of notes found on an EA diagram itself.

PeterStange:
This pull request fixes the issue #450. It adds the export of notes that are found for the diagram itself.
Without it, all notes found on elements used by the diagram are exported only.
As it requires to start the notes on elements with {adoc: to be exported, it is the same for diagram notes.
If the tag is not found at the beginning of a note, the note is not going to be exported.

There was no need to update the documentation of exportEA.

I created this pull request as a representative of AndreaMcls.


rdmueller:
thanx!

PR #514, opened 2021-01-13
Add support of multiple EA project locations in parallel 

PeterStange:
Both project accessed by database connection and by local EA project files should work in parallel.

First implementation was exclusively database or local files. 
Got several complains because of this and therefore I checked parallel support of both types and there is no issue supporting this.
Avoiding collisions in case of multiple projects in parallel should be managed by the projects itself.
Usually differently named root package should be sufficient.

I adopted documentation for exportEA script.

Test and feedback were positive until now.


rdmueller:
thanx for this PR - sorry that it took so long to merge it.

Issue #529, opened 2021-02-10
exportEA diagramAttributes are not exported correctly


PeterStange:
**Describe the bug**
exportEA configuration contains the diagramAttributes like:
diagramAttributes = "diagram name: %DIAGRAM_NAME%; author=%DIAGRAM_AUTHOR%; last modified: %DIAGRAM_MODIFIED%" 

Run 'gradle exportEA'
After export script finished, all created attributes files contain the same content, but not the attributes of the related diagram.

After analysis, the root cause of this behaviour is an implementation bug. All diagrams get the attributes of the first diagram. A proper fix is already available and will be offered as a pull request immediately.

PeterStange:
Solved with #530

PR #530, opened 2021-02-10
Fix writing of diagram attributes using exportEA

PeterStange:
This pull request fixes bug #529.

The script exportEA supports export of diagram attributes like author, name and modified date.
Because of a bug in the script, all diagrams gets the same attributes - the attributes of the first read diagram.
Sorry for this. I have to be more careful with merges in future.

Additionally, I introduced the const 'ForWriting'. By this, the files written are newly created and existing content will be removed. Previously used 'ForAppending' is not causing direct errors, but is not the intended functionality. Furthermore, in case same name is used for multiple diagrams, the last exported diagram is available only, but because of appending the attributes to file, the attributes of all diagrams with the same name were added.

### DIFF
--- a/scripts/exportEAP.vbs
+++ b/scripts/exportEAP.vbs
@@ -10,6 +10,7 @@
     Dim projectInterface 'As EA.Project
     
     Const   ForAppending = 8
+    Const   ForWriting = 2
     
     ' Helper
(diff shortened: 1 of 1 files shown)


rdmueller:
cool! Thanx for this PR!

Issue #570, opened 2021-04-29
exportEA hang after all exports are done since EA v15.2


PeterStange:
**Describe the bug**
We installed the version 15.2 of Enterprise Architect. Since this time calling exportEA never ends. It stops after all exports are done. After manual termination of the background process (Enterprise Architect), the script ends successfully.

**Expected behavior**
The script shall end automatically.

**Configuration**
 - OS: Windows
 - Enterprise Architect v15.2

PeterStange:
Potentially fixed with pull request #571

PeterStange:
With merge #571 is fixed

PR #571, opened 2021-04-29
Call Exit() of repository class to terminate the EA process on demand.

PeterStange:
As described in bug #570, with newer version of Enterprise Architect the script exportEA hangs while waiting for termination of the background process. To trigger this, the Repository API offers the Exit() method. 
Reference text: "Notes: Shuts down Enterprise Architect immediately. Used by .NET programmers where the garbage collector does not immediately release all referenced COM objects."

After adding this call, the script automatically proceeds until it ends successfully. It is tested with online connections and local eap files (even with combinations of both).
As this bug prevents us from periodical, automatic document generation, this pull request is from importance for us.

### DIFF
--- a/scripts/exportEAP.vbs
+++ b/scripts/exportEAP.vbs
@@ -366,6 +366,8 @@
         End If
       End If
       EAapp.Repository.CloseFile()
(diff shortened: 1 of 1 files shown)


rdmueller:
thanx!

Issue #486, opened 2020-10-27
test Failing when using the gradle wrapper


AndreaMcls:
**Problem Description**
The test tasks are failing when executing with the gradle wrapper.
The same tests are not failing when a local gradle installation eg. version 6.4 is used.
All other tasks eg: generateHTML, exportEA, pulishToConfluence, etc..  are working as expected.

**To Reproduce**
1. cd to doctoolchain rootfolder
2. ./gradlew test --info
3. All tests are failing
4. Error Message:
`  > Configure project :
    Evaluating root project 'docToolchain' using build file 'D:\test\docToolchain\build.gradle'.

    FAILURE: Build failed with an exception.

    * Where:
    Build file 'D:\test\docToolchain\build.gradle' line: 6

    * What went wrong:
    Plugin [id: 'org.asciidoctor.convert', version: '2.4.0'] was not found in any of the following sources:

    - Gradle Core Plugins (plugin is not in 'org.gradle' namespace)
    - Plugin Repositories (could not resolve plugin artifact 'org.asciidoctor.convert:org.asciidoctor.convert.gradle.plugin:2.4.0')
      Searched in the following repositories:
        Gradle Central Plugin Repository

    * Try:
    Run with --stacktrace option to get the stack trace. Run with --debug option to get more log output. Run with --scan to get full insights.

    * Get more help at https://help.gradle.org

    BUILD FAILED in 5s
        at org.gradle.testkit.runner.internal.DefaultGradleRunner$2.execute(DefaultGradleRunner.java:255)
        at org.gradle.testkit.runner.internal.DefaultGradleRunner$2.execute(DefaultGradleRunner.java:251)
        at org.gradle.testkit.runner.internal.DefaultGradleRunner.run(DefaultGradleRunner.java:324)
        at org.gradle.testkit.runner.internal.DefaultGradleRunner.build(DefaultGradleRunner.java:251)
        at docToolchain.ExportChangeLogSpec.test creation of log file(ExportChangeLogSpec.groovy:19)
`

**Expected behavior**
test are executed and passed. Tests are passed when using a manual installed gradle (for example v6.4)

**Configuration**
 - OS:  WIndows
 - Relevant SDK Versions JDK 14

**Additional context**

**Find a good headline**
test Failing when using the gradle wrapper

Example:
exportContributors throws exception when ...

ehmkah:
[...]

PR #572, opened 2021-04-30
Issue486 proxy setting test

AndreaMcls:
Document solution of issue #486


rdmueller:
Thanx! this is great.
And welcome as a contributor!

Issue #612, opened 2021-07-29
maven credentials


Hellmy:
**Is your feature request related to a problem? Please describe.**
I want to use a corporate maven-proxy which requires authentication. I can change the url in the gradle.properties but how can the authentication be done?

**Describe the solution you'd like**
We could introduce two other properties (mavenUsername and mavenPassword) but I'm not sure if it works with empty properties
```
[..]
    repositories {
        maven {
            credentials {
                username mavenUsername
                password mavenPassword
            }
            url mavenRepository
        }
    }
[..]
```

**Describe alternatives you've considered**
The user has to configure the mavenRepository in the gradle-home settings. But that collide with the mavenRepository configuration of the project

rdmueller:
Great PR. Sorry, I still need some time to check the changes

rdmueller:
maybe we can use Groovy code to only set the credentials when they are defined. This should solve the problem:

https://stackoverflow.com/questions/43901251/setting-credentials-in-gradle-via-maven-publish

jmini:
I am interested in re-doing PR https://github.com/docToolchain/docToolchain/pull/626 in the `ng` branch. 

I will file a PR soon.

PR #626, opened 2021-09-10
Add maven credentials #612

Hellmy:
Tested it with credentials section and empty credentials with the default mavenRepository - it just worked, so no if-groovy block required.


rdmueller:
looks good. thanx!

PR #628, opened 2021-09-17
Export ea add check for existing files

Rosi2143:
Add check if an exported EA-Diagram already exists

If a diagram with the same name exists multiple times it
is potentially overwritten via the task exportEA.

To let the user decide if this is intended an additional
parameter is added. This allows or prevents that a
diagram is overwritten.

additionalOptions = "KeepFirstDiagram"

Signed-off-by: rmi2hi <michael.rossner@de.bosch.com>


Rosi2143:
One more change done to improve string check.

Issue #506, opened 2021-01-04
generateHTML does not copy images directory


softmetz:
The task generateHTML does not copy over the contents of the `taskInputsDirs`, which leads to images missing in the output.

Adding `:data-uri:` in `config.adoc` has no effect either.

On contrary PDF files created with generatePDF carry all images. 
The same is true for `generateHTML` inside fresh checkout of docToolchain where the `images` is copied below build.

**To Reproduce**
1. Add as submodule at `docToolchain` as in https://docs-as-co.de/getstarted/tutorial2
2. extract ZIP of arc42 EN in src/arc42
3. Adapt Config.groovy accordingly
4. Run ./gradlew generateHTML
5. Look for images folder in `build/docs/html5` -> Its missing

**Expected behavior**
The folder `images` exists below `build/docs/html5`

**Screenshots**
```
build
└── docs
    ├── html5
    │   └── arc42
    │       └── omkring-architecture-description.html
    └── pdf
        └── arc42
            └── omkring-architecture-description.pdf
```

```
src/
├── arc42
│   ├── images
│   │   ├── arc42-logo.png
│   │   ├── arc42-logo.png.license
│   │   ├── omkring-logo.svg
│   │   └── omkring-logo.svg.license
│   ├── omkring-architecture-description.adoc
│   └── src
│       ├── 01_introduction_and_goals.adoc
│       ├── 02_architecture_constraints.adoc
│       ├── 03_system_scope_and_context.adoc
│       ├── 04_solution_strategy.adoc
│       ├── 05_building_block_view.adoc
│       ├── 06_runtime_view.adoc
│       ├── 07_deployment_view.adoc
│       ├── 08_concepts.adoc
│       ├── 09_design_decisions.adoc
│       ├── 10_quality_scenarios.adoc
│       ├── 11_technical_risks.adoc
│       ├── 12_glossary.adoc
│       ├── about-arc42.adoc
│       └── config.adoc
└── pdfTheme
    └── custom-theme.yml
```

Config.groovy:
```groovy
outputPath = 'build/docs'

// Path where the docToolchain will search for the input files.
// This path is appended to the docDir property specified in gradle.properties
// or in the command line, and therefore must be relative to it.
inputPath = 'src'

inputFiles = [
              [file: 'arc42/omkring-architecture-description.adoc',            formats: ['html','pdf']],
             ]

taskInputsDirs = ["${inputPath}/arc42/images"]

taskInputsFiles = []
```

build.gradle:
[...]

PR #686, opened 2021-11-03
Add resourceDirs option #506

Hellmy:
Works for generateHTML and generateSite 

should be checked if applicable for generateSite but seemed right to implement it there as well

Targets #506


rdmueller:
sorry, that it took me so long to check this PR...

PR #708, opened 2021-12-03
Correct confluence password parameter name

sideisra:
I had problems using the publish to confluence feature because i could not figure out how to pass the credentials. I finally found that i misstyped the password parameter. The documentation says that i can use "-PconfluencePassword" but it's actually "-PconfluencePass".

I'm a little confused about the branches. Is "ng" the right branch to fix this? Should i fix it on other branches to?


rdmueller:
yes, "ng" is the right branch. It stands for "next generation".
Thanx for your fix!

PR #711, opened 2021-12-06
Update asciidoc2confluence.groovy to handle attachements without hash

Rosi2143:
Somebody managed to upload attachments without a hash, that broke the job publishToConfluence.

This change checks first if a hash is available before extracting it. Otherwise hash will be set to 0 - which is at least different to the real hash.

Signed-off-by: Michael Roßner Schrott.Micha@web.de

### All Submissions:

  * yes
  * no

### DIFF
--- a/changelog.adoc
+++ b/changelog.adoc
@@ -7,6 +7,10 @@ and this project tries to adhere to https://semver.org/spec/v2.0.0.html[Semantic
 
 == Unreleased
 
+=== fixes
+2021-12-06::
+* https://github.com/docToolchain/docToolchain/pull/711[#712 - publishToConfluence fails when no hash is available for an uploaded image]
+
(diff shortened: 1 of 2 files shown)


Rosi2143:
Problem with the Jenkins job seems to be here:

see e.g. https://github.com/docToolchain/docToolchain/runs/4436533630?check_suite_focus=true

> Task :generateHTML NO-SOURCE
> Task :htmlSanityCheck FAILED

There is no HTML generated and that's why later on the SanityCheck fails with

> Directory '/home/runner/work/docToolchain/docToolchain/build/docs/html5' specified for property 'sourceDir' does not exist

rdmueller:
thanx for this PR! I will deactivate the netlify build since we switched to github actions

PR #756, opened 2022-01-23
exportEA: Add two more items for export

Rosi2143:
- Version
- Type

Signed-off-by: Schrotti <Schrott.Micha@web.de>

### All Submissions:

* [y] Did you update the `changelog.adoc`?
* [y] Does your PR affect the documentation?
* [y] If yes, did you update the documentation or create an issue for updating it?

PR #757, opened 2022-01-23
CI: Fix problem in ci-script

Rosi2143:
Signed-off-by: Schrotti <Schrott.Micha@web.de>

### All Submissions:

* [y] Did you update the `changelog.adoc`?

### DIFF
--- a/.ci.sh
+++ b/.ci.sh
@@ -106,7 +106,9 @@ create_doc () {
 #    cp -r build/docs/report/htmlchecks/. docs/htmlchecks/.
 
   else
+    echo ">>> exportMarkdown"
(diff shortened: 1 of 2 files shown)


rdmueller:
thanx. is this part of the script still in use? I thought it was ok to comment it out since it only runs on the old branches.

Rosi2143:
I don't know, but my #756 failed due to this missing change in .cl.sh.

We can remove the "else" branch completely. I just wasn't sure.

rdmueller:
yes, I think we can remove it.

rdmueller:
btw: I still ahve the problem that this script is triggered from PRs an fails...

PR #822, opened 2022-03-09
explain some more of the inner magic of the menu

rdmueller:


src/docs/015_tasks/03_task_generateSite.adoc
```diff
@@ -84,6 +84,48 @@ For `:jbake-type: page_toc`, you need this line to generate the toc.
...
+[[config]]
+== Configuration
+
+The configuration follows the __convention over configuration__ approach. 
+If you follow the conventions, you don't have to configure anything.
+But if you want to, you can override the convention behaviour with a configuration.
+
+=== Menu
+
+The `:jbake-menu:` is only the code for the menu entry to be created.
```
JodyWinter, line 96: I'm not sure what this means. Are you saying that the j:bake-menu is the only code required for the menu entry to be created?

rdmueller: oh, good question. 

there is a code or token for the menu like menu1, menu2 etc. Just some text to identify the menu item, not the text to be rendered.

This code or token is derived from the folder name or directly taken from the :jbake-menu: attribute.

It is then translated thorugh the 'menu' config to the rendered string:

menu = [menu1: "my first menu", menu2: "my second menu"]

this is necessary, because not all characters are allowed in folder names.

But if no translation is given, the code or token will be rendered instead.

Does this clarify tings?

JodyWinter: **Is this correct? I don't fully understand this - a bit too developer-y for me, but let me know if I am close.**

You can use a code such as menu1, menu2, menu3... etc. as the `:jbake-menu:` attribute to create the menu structure instead of the full menu folder name. The code is translated by the `menu` config to the rendered string. This is useful because not all characters are allowed in folder names.

The code and folder name are mapped using `menu` in the configuration (`microsite`-Section) in the following way:

menu = [code1: 'title1', code2: 'title2', code3: '-']

rdmueller: great. I think this is as close as it can get :-)

Issue #692, opened 2021-11-10
generateSite: navigation tree instead of navigation list on the left


jmini:
Given a structure like this:

```
src/docs/
├── ...
├── 30_tri
│   ├── 10_ipsum.adoc
│   └── 20_advanced
│       ├── 10_topic-A.adoc
│       └── 20_topic-B.adoc
├── ...

```

`generateSite` seems to ignore the fact that some pages were put in a sub-folder.
In the navigation tree on the left, everything is flat:

<img width="769" alt="docToolchain output of the generateSite command" src="https://user-images.githubusercontent.com/1222165/141083180-f483bfb6-05a7-4cbe-9111-a801695d91a5.png">

**Describe the solution you'd like**
I would like to see a tree of pages (instead of a list) where folders of the source files are considered.

Concretely in this example, I would have expected `topic-A` and `topic-B` to be under an `advanced` node:

<img width="295" alt="Screenshot 2021-11-10 at 10 04 31" src="https://user-images.githubusercontent.com/1222165/141083309-086434f7-faf9-49fd-b14b-b98c90d2c83f.png">

rdmueller:
Yes! I also would love to see this feature 😀 
We will have soon an open source Hackergarten. I think this would perfectly fit...

jmini:
@rdmueller sorry to come back on this a long time after this was posted. 

Do you have any idea/pointer how this would start?

rdmueller:
there are two ways we can solve this:

### quick and dirty

in some projects, I used to create such a navigation just as list of links:

```
* xref:first.adoc[1. first]
** xref:second.adoc[1.1 second]
* xref:third.adoc[2. third]
```

with the right template, you can replace the standard menu with this manually curated list.

### automated

This is where the meta-attribute `:jbake-menu:` is set:
https://github.com/docToolchain/docToolchain/blob/ng/scripts/generateSite.gradle#L246

...and this is where the `:jbake-title:` is set:
https://github.com/docToolchain/docToolchain/blob/ng/scripts/generateSite.gradle#L287

both together make up the menu: `:jbake-menu:` is the menu entry on top of the page, `:jbake-title:` is the name of the link to the left.

I think we need something more struktured here... if we would have maybe another meta-attribute like `:jbake-deepmenu:` which gets the folder structure assigned, we could use it in

https://github.com/docToolchain/docToolchain/blob/ng/src/site/groovy/menu.groovy

To build up an array which is then used in 
[...]

PR #864, opened 2022-06-29
Add tests for 'src/site/groovy/menu.groovy'

jmini:
While working on #692, I noticed there is no test class for the `src/site/groovy/menu.groovy` script.

This PR is an attempt to write such a test class and to collect feedback (to know if I go in a good direction or not).

This test class is supposed to be extended when working on #692.

---

In my opinion there is no need to document this in the `changelog.adoc` file, since there is no change to the code. If you have an other opinion, I am happy to add an entry.

### DIFF
--- a/src/test/groovy/site/MenuSpec.groovy
+++ b/src/test/groovy/site/MenuSpec.groovy
@@ -1,0 +1,56 @@
+package docToolchain
+import spock.lang.*
+import org.gradle.testkit.runner.GradleRunner
+import static org.gradle.testkit.runner.TaskOutcome.*
+
+class MenuSpec extends Specification {
+
+    class ContentFixture {
+        def menu
+        def entriesMap
+        def newEntries
+    }
+
+    void 'test empty published content'() {
+        given: 'empty published_content'
+            Binding binding = new Binding()
+            binding.published_content = []
+        when: 'run the `menu.groovy` script'
+            runMenuScript(binding)
+        then: 'arrays are empty'
+            binding.content.menu == [:]
+            binding.content.entriesMap == [:]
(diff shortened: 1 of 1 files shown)


src/test/groovy/site/MenuSpec.groovy
```diff
@@ -0,0 +1,56 @@
...
+            binding.content.menu == [:]
+            binding.content.entriesMap == [:]
+            binding.content.newEntries == []
+    }
+
+    void 'test with published content'() {
+        given: '3 pages in published_content'
+            Binding binding = new Binding()
+            binding.published_content = [
+                ['jbake-menu': 'foo', 'jbake-title': 'Lorem Ipsum', 'jbake-order': 10, uri : 'foo/10_lorem-ipsum.html'],
```
jmini, line 30: The `order` seems to be of type `String` in the array in reality

rdmueller: yes, I think I already fixed this somewhere...

rdmueller:
I really appreciate these tests.
To be honest, I struggle a bit with creating automated tests for docToolchain because they are not easy to create. In this case, I think you found a good way. Thanx for this contribution!

rdmueller:
I love the approach you take here. I wonder if we can use it also for other code.

src/test/groovy/site/MenuSpec.groovy
jmini: Maybe you think of https://github.com/docToolchain/docToolchain/pull/803

But internally in this array the order is a `String`. I think it is converted to `int` later

PR #908, opened 2022-08-25
Tutorial publish to confluence

rdmueller:


rdmueller:
@JodyWinter I tried to come up with a new tutorial (because a user had a problem).
If you find the time, it would be great if you could do a quick review. 
If you lack the time, maybe we can just open an issue and proceed with the PR?

rdmueller:
thanx for the edits!

PR #951, opened 2022-10-26
Improve DTC_PROJECT_BRANCH management

### LINKED ISSUE #950: Branch name is not correctly set when exectued in a CI
**Describe the bug**
When running on GitHub action (could be the same with other ci system as well) it seems that the branch name is not available.

Probably the `git` command is not working here:
https://github.com/docToolchain/docToolchain/blob/91404b6a99c339c67e9c067cee55f592f4c37d68/dtcw#L19

Which set a wrong value to `DTC_PROJECT_BRANCH`.

Which leads to the `gitRepoUrl`  to be wrong here:
https://github.com/docToolchain/docToolchain/blob/91404b6a99c339c67e9c067cee55f592f4c37d68/docToolchainConfig.groovy#L101-L103

**To Reproduce**
1. Have a setup with continuous publication to `gh-pages` using a Github action (see the [ci-cd-demo](https://gitlab.com/doctoolchain/ci-cd-demo/) project)
2. Trigger the github action to run `generateSite` and to push the result
4. See error: reproduced in https://github.com/jmini/dtc-experiments/issues/3

**Expected behavior**
I think the `dtcw` command should set only `DTC_PROJECT_BRANCH` when it is not already set.

This way we could set the branch name value coming from the CI pipeline directly and it would not be replaced by `dtcw`.

jmini:
Fixes #950

* This sets `DTC_PROJECT_BRANCH` only if it was not already set from an other process
* It propagates the `DTC_PROJECT_BRANCH` value to the process running inside docker.

### DIFF
--- a/changelog.adoc
+++ b/changelog.adoc
@@ -12,6 +12,7 @@ and this project tries to adhere to https://semver.org/spec/v2.0.0.html[Semantic
 * fixed ExportMarkdownSpec
 * fixed GenerateDeckSpec
 * allow numeric ancestorIds for confluence export
+* https://github.com/docToolchain/docToolchain/pull/951[#951 Improve DTC_PROJECT_BRANCH management]
 
 === added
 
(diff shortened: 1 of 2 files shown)


jmini:
Follow up PR for the `ci-cd-demo` repository: https://gitlab.com/doctoolchain/ci-cd-demo/-/merge_requests/2

rdmueller:
thanx!

PR #993, opened 2023-01-10
Update AsciiDocBasics.gradle

Rosi2143:
Add check if all files in sourceFile<Type> exist. Otherwise throw an exception.

### All Submissions:

* [n] Did you update the `changelog.adoc`?

Signed-off-by: Rosi2143 [Schrott.Micha@web.de](mailto:Schrott.Micha@web.de)

### DIFF
--- a/Config.groovy
+++ b/Config.groovy
@@ -8,7 +8,7 @@ inputPath = 'src/docs';
 
 
 inputFiles = [
-        [file: 'manual.adoc',       formats: ['html','pdf']],
+        [file: 'manual_test_script.adoc',       formats: ['html','pdf']],
         /** inputFiles **/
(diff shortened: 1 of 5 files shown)


rdmueller:
thanx for this PR. I wonder if we can design the message part of the exception a bit more helpful. Something which tells the user where to look for the problem?

Rosi2143:
> thanx for this PR. I wonder if we can design the message part of the exception a bit more helpful. Something which tells the user where to look for the problem?

Hi Ralf,

sure - but let me fix the failing tests first - there are a couple of files mentioned in the config.groovy that do not exist :-).

I'll rewrite the error message once all tests pass.

PR #997, opened 2023-01-16
940 preamble heading

aklemp:
### All Submissions:

### Your first submission


rdmueller:
thanx for this PR.

Do I read the code correct that this is not a breaking change - old configs will still work?

aklemp:
I removed the `preambleTitle` configuration and to get the original behavior, the user has to change the first level heading to the value that was previously in that config value. However, the tool doesn't explode but stops with a (helpful) migration message. I have not implemented this as a fallback like "if config is set, use it and the first level heading otherwise" because that would also not keep the original `arc42` default.

rdmueller:
ok, so we have  a breaking change. Good to know.
Thanx for the clean implementation of this feature!

PR #998, opened 2023-01-19
Update collectIncludes.gradle - generated pathes consider `scanDir` folder

Rosi2143:
if scanDir is used (e.g. "./src/arc42" the generated path is short by this setting.

To make the files accessible again we need to add the (original) scanDir  again to the generated path.

Signed-off-by: Rosi2143 Schrott.Micha@web.de


rdmueller:
thanx!!!

PR #1046, opened 2023-02-21
1045 dtcw without help

### LINKED ISSUE #1045: Calling the wrapper script without any argument does not provide helpful output
This is a regression already reported in #1043.

Calling the wrapper script without any argument does not provide helpful output. The the wrapper script of version v2.1.0 provided a good description what the user should do next. 

**To Reproduce**
1. Install the wrapper script
```
~/tmp 
➜ mkdir test && cd test

~/tmp/test 
➜ curl -sLo dtcw doctoolchain.github.io/dtcw

~/tmp/test 
➜ chmod +x dtcw 
```
2. Call the wrapper script with no arguments
```
~/tmp/test 
➜ ./dtcw

dtcw - docToolchain wrapper V0.36
docToolchain V2.2.0
docToolchain as CLI available
No valid argument(s) supplied
```

**Expected behavior**
The user is provided with helpful information what they should do next. In previous releases we had following output.
```
./dtcw 

dtcw - docToolchain wrapper V0.36
docToolchain V2.1.0
docToolchain as CLI available
local java JDK found
use /home/max/.doctoolchain/jdk as JDK

Usage: ./dtcw [option...] [task...]

You can use the same options and tasks as in underlying gradle.
Use "./dtcw tasks --group doctoolchain" to see available tasks.
Use "local", "sdk" or "docker" as first argument to force the use of a local, sdkman or docker install.

Examples:

  Download and install arc42 Template:
    ./dtcw downloadTemplate

  Generate PDF:
    ./dtcw generatePDF

  Generate HTML:
    ./dtcw generateHTML

  Publish HTML to Confluence:
    ./dtcw publishToConfluence

  get more documentation at https://doctoolchain.github.io
```

**Screenshots**

**Configuration**
 - docToolchain version: v2.20
 - OS: Linux
 - Relevant SDK Versions: the one provided with `getJava`

rdmueller:
this should fix the issue.

### DIFF
--- a/dtcw
+++ b/dtcw
@@ -41,7 +41,7 @@ export DTC_HEADLESS
 
 DTC_OPTS="${DTC_OPTS:-} -PmainConfigFile=${MAIN_CONFIG_FILE} --warning-mode=none --no-daemon "
 
-echo "dtcw - docToolchain wrapper V0.36"
+echo "dtcw - docToolchain wrapper V0.37"
 echo "docToolchain V${DTC_VERSION}"
 
 # check if CLI, docker or sdkman are installed
--- a/dtcw
+++ b/dtcw
@@ -70,8 +70,30 @@ if [ -d "${HOME}/.doctoolchain/docToolchain-${DTC_VERSION}/." ]; then
     echo "home folder exists"
     homefolder=true
 fi
-if [ $# -eq 0 ]; then
-    echo "No valid argument(s) supplied"
+if [[ $# -lt 1 ]]; then
+    echo '
+Usage: ./dtcw [option...] [task...]
+
+You can use the same options and tasks as in underlying gradle.
+Use "./dtcw tasks --group doctoolchain" to see available tasks.
+Use "local", "sdk" or "docker" as first argument to force the use of a local, sdkman or docker install.
+
+Examples:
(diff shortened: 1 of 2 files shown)


rdmueller:
@mh182 since you are currently so active helping - would you like to review my change before we merge it?

mh182 (approved):
Looks good to me.

mh182:
I"m not used to work/cooperate on github. I rebased and merged the PR. But the commit doesn't contain a reference to the original ticket number.
It is not a good practice to integrate commits which don't reference to any ticket number.

rdmueller:
I opened the new branch from within the issue. This auto-generates a branch name which starts with the issue number. This seems to be enough reference for github. But to be honest, I also switched to this workflow only aproc 10 issues ago. So let's see how this works...

rdmueller:
Bte: I also try to squash commits and use conventional commit messages for the PR in order to try to keep the history cleaner. But this is only on a best effort basis and also just since two weeks 😁

mh182:
The github interface didn't provide me a rabes+squash option (what I was searching for). 
I would prefer you are merging the issues until I get more used to the HG workflow here.

PR #1037, opened 2023-02-21
added (ADR): ADR01-Structurizr

rdmueller:

Issue #1016, opened 2023-02-09
Export diagrams from software architecture models provided in Structurizr DSL format


mh182:
Provide means to export diagrams from a software architecture model provided as [Structurizr DSL](https://structurizr.com/dsl).

**Describe the solution you'd like**
Add a new `gradle` task `exportStructurizr` which uses the https://github.com/structurizr/cli to export the diagrams from a DSL file located on the file system (i.e. in the `doc/src`). 

**Describe alternatives you've considered**
The Structurizr CLI contains the needed functionalities to export the diagrams in various formats. I could not think of an alternative solution

**Additional context**
I tired to implement a proof of concept using the Structurizr plugin provided from https://github.com/jakzal/gradle-structurizr-cli. See https://github.com/mh182/docToolchain/tree/exportStructurizr 

But I lack the know-how in the different technologies used by `docToolchain` (Gradle, Kotlin, Java) and hit a block. I have no idea how to set the properties needed for the Gradle plugin. I even don't know if it makes sense to use the Gradle plugin or there is an easier way to use  Structurizr CLI directly.

mh182:
Attached a demo environment ([doctools-exportStructurizr-example.zip](https://github.com/docToolchain/docToolchain/files/10697192/doctools-exportStructurizr-example.zip)) with examples to test the feature.

```
# Setup dev environment
➜ cd $HOME/.doctoolchain
➜ git clone -b exportStructurizr https://github.com/mh182/docToolchain.git docToolchain-2.1.0-exportStructurizr

# Setup example environment
➜ wget https://github.com/docToolchain/docToolchain/files/10697192/doctools-exportStructurizr-example.zip
➜ unzip doctools-exportStructurizr-example.zip
➜ cd doctools-exportStructurizr-example
➜  ./dtcw local task
# Export the diagrams - a hack since my exportStructurzr task doesn't wotk yet
➜ ./dtcw local structurizrCliExport
➜ ./dtcw local generateHTML
```

If anyone could point me in the right direction how to integrate Structurizr CLI I would be willed to finish and polish this feature for integration.

rdmueller:
hey, this already looks great!

rdmueller:
you already got quite far.
I tried to replicate this with the bitsmuggler example. 

To develop such new tasks, you can now create tasks in your own project (dtc v2.2.0).
Execute `./dtcw createTask` to create a dummy task.

[...]

PR #1056, opened 2023-02-24
Add new task exportStructurizr

mh182:
I think we are almost ready for integration. 

It's a draft PR since we have to address the following issues:
* the test case is still failing (and I have no idea why because the diagrams are created)
* the task documentation is not finished (I don't know how to generate the HTML output locally to test it)
* currently, the diagrams are created each time a `generate*` task is called. Is there a way to generate the files only if the workspace file changed?
* how do we proceed with #1037? I would like to finish that one before we integrate the code for this ticket.


rdmueller:
regarding the error:

```
docToolchain.ExportStructurizrSpec > test conversion of Structurizr DSL file to diagrams FAILED
    Condition not satisfied:

    result.task(":exportStructurizr").outcome == SUCCESS
    |      |                          |       |
    |      |                          |       false
    |      |                          UP_TO_DATE
    |      :exportStructurizr=UP_TO_DATE
    <org.gradle.testkit.runner.internal.FeatureCheckBuildResult@deebb6d delegateBuildResult=org.gradle.testkit.runner.internal.DefaultBuildResult@2f6150df outputFeatureCheck=org.gradle.testkit.runner.internal.feature.BuildResultOutputFeatureCheck@18ff5d11>
        at docToolchain.ExportStructurizrSpec.test conversion of Structurizr DSL file to diagrams(ExportStructurizrSpec.groovy:28)
```

The task responds with "UP_TO_DATE" instead of "SUCCESS". I guess this is because the diagrams are already there?

rdmueller:
ok. the test doesn't fail locally. I currently try to update the github action to store the test results as artifact.

rdmueller:
I created a PR to get the test report: https://github.com/mh182/docToolchain/pull/1

rdmueller:
Here is the docs preview: https://deploy-preview-1056--dtc-docs-preview.netlify.app/015_tasks/03_task_exportstructurizr

(last task of automated tasks in a PR)

rdmueller:
I was on the wrong branch. Now I also get the error locally

rdmueller:
I think I found the problem:

Gradle Tasks have phases. Out scripts have to run in the `doLast` phase of the task.

rdmueller:
> currently, the diagrams are created each time a generate* task is called. Is there a way to generate the files only if the workspace file changed?

yes, there are ways, but they create more problems than they solve.

Shall we merge?

mh182:
I started to improve the test coverage. Let me check your changes when I'm at home.

I would like to finish the task documentation before squash merging it into `ng`.

Before the merge I would like to look into how the workflow may change if an author uses Structurizr Light (SL). I just want to make sure we can integrate using SL without breaking the configuration.

I will drop you a line when I thin the feature branch is complete.

mh182:
> Gradle Tasks have phases. Out scripts have to run in the `doLast` phase of the task.

Could you explain to me why do those tasks have to run in the `doLast` phase? 

As far as I understood, `doFirst` and `doLast` are optional, and you can use them to perform any actions you want before or after the main action of a task (exporting the diagrams).

The most obvious purpose for `doFirst`/`doLast` is code reuse. For example, when using a Gradle built-in tasks like copy, and you wanted to do something before the copy happens, or afterward on cleaning up resources.

rdmueller:
> Could you explain to me why do those tasks have to run in the `doLast` phase?

good question. 

the `doFirst` phase seems to be a preparation phase which runs for all tasks always. In this phase, you can configure dependencies of the task and check if files have changed.

the change from nothing to `doLast` did at least fix our problem.

don't know which phase it is when nothing is specified.

as you can see, my gradle knowledge is a bit outdated.

mh182:
I addressed your feedback and rebased on the `ng` branch. From my part, I think it is ready to be integrated with a squash merge.

Changing the task to run its commands with `doLast` had a visible effect. The diagrams are not exported automatically when I call `generateHTML` anymore. Now I have to execute `./dtcw exportStructurizr` manually before executing a `generate*` task. Is this behavior wanted?

I have one minor issue with the task documentation. There is a Structurizr workspace file in `src/test/testData/structurizr/workspace.dsl` for the test cases. I tried for 2 hours, but was not able to include the code in the task documentation. At the end, I copy/pasted the code.

rdmueller:
> The diagrams are not exported automatically when I call generateHTML anymore. Now I have to execute ./dtcw exportStructurizr manually before executing a generate* task. 
> Is this behavior wanted?

yes, I think this is what we need. My feeling is that before, the exportStruturizr was _always_ executed. Even when not needed.
We could create a dependency, but my experience is that it makes sense to run the tasks in an explicit order.

You can also execute `./dtcw exportStructurizr generateHTML` in one go.

Do you think we should change this?

rdmueller:
> I have one minor issue with the task documentation.

let me give it a try...

rdmueller:
this part?

```
[source]
----
include::{projectRootDir}/src/test/testData/structurizr/workspace.dsl[]
----
```

the `{projectRootDir}` should work here...

rdmueller:
I think you are right. We should add a `dependsOn` for `generateHTML`, `generatePDF` and `generateSite`. The `exportStructurizr` will be skipped if there is no DSL, so it will only be active if it is used.

mh182:
> this part?
> 
> ```
> [source]
> ----
> include::{projectRootDir}/src/test/testData/structurizr/workspace.dsl[]
> ----
> ```

Yes, that was the part I was referring to. I tried absolute path with `projectRootDir` and relative path with `../../`. 
I just realized my mistake was the missing `[]` after the file name. I haven't used Asciidoc a lot.

I will try it out

mh182:
> I think you are right. We should add a dependsOn for generateHTML, generatePDF and generateSite. The exportStructurizr will be skipped if there is no DSL, so it will only be active if it is used.

I think the most important aspect of a software tooling is consistency. If the excel/Visio/openApi files are created automatically when calling generate* we should stick to it.

The main problem is the runtime is going to add up when adding more exporters. On the other side I have no experience how fast asciidoctor is to convert an average file. 
If it takes 30 seconds to convert a document the 0.5 seconds for the check is negligible.
The materialize part on the first call wouldn't hurt much either because imho most users will use the container image anyway.

So my vote goes for consistency.

mh182:
I replaced the code block with the `include` and made minor changes to the documentation in the configuration files.

I noticed that we would have to update the picture provided in [At a Glance](http://doctoolchain.org/docToolchain/v2.0.x/015_tasks/03_tasks.html#_at_a_glance).

Regarding the task dependencies - I just checked the `generateHTML`. As far as I can tell, the other file types like Visio, Excel etc. are also not generated automatically. So let's be consistent and the person using the tooling should know when to call the relevant `export*` tasks.

rdmueller:
wow. I just used gitpod.io to update the diagram within the browser. Worked like a charm. I am really impressed.

https://deploy-preview-1056--dtc-docs-preview.netlify.app/015_tasks/03_tasks.html

So, I guess we can squash and merge?

mh182:
> wow. I just used gitpod.io to update the diagram within the browser. Worked like a charm. I am really impressed.
> 
> https://deploy-preview-1056--dtc-docs-preview.netlify.app/015_tasks/03_tasks.html
> 
> So, I guess we can squash and merge?

GitHub is still a mystery to me. How are you able to add a change on a PR which comes from a branch in my repo?

Your change only contains the PNG file. But where are the changes in the draw.io file?

About the merge please check my https://github.com/docToolchain/docToolchain/issues/1016#issuecomment-1445115140.

rdmueller:
> How are you able to add a change on a PR which comes from a branch in my repo?

When you start a PR, there is a default option to allow edits on your branch. Quite helpful for the code review

> Your change only contains the PNG file. But where are the changes in the draw.io file?

The PNG file contains the draw.io source in the meta data. That is quite cool :-)

mh182:
> > How are you able to add a change on a PR which comes from a branch in my repo?
> 
> When you start a PR, there is a default option to allow edits on your branch. Quite helpful for the code review

I didn't know that. Thx. At least I was able to find your commit ;-).

> > Your change only contains the PNG file. But where are the changes in the draw.io file?
> 
> The PNG file contains the draw.io source in the meta data. That is quite cool :-)

Wow, this is really cool. 
But then we should delete the old, outdated file `src/docs/015_tasks/overview.dio.xml.drawio` which I was trying to change and failed miserabely.

rdmueller:
Thanx. I just removed the outdated files.

Squash end Merge?

mh182:
> Thanx. I just removed the outdated files.
> 
> Squash end Merge?

No, I want to make the minimal changes regarding the configuration as described in https://github.com/docToolchain/docToolchain/issues/1016#issuecomment-1445115140.

The changes I want to incorporate before merge are:
- The workspace file is defined with `structurizr.workspace.path` and `structurizr.workspace.filename`. 
- Abort in case no workspace file can be found (instead of skipping export), since `exportStructurizr` is called explicitly.

mh182:
I think it is ready for integration now. I addressed the configuration settings and updated the documentation

If this looks good to you, please integrate it with a squash merge.

rdmueller:
thanx for the work on this feature. I really enjoyed it!

Issue #1062, opened 2023-02-26
Local installation of docToolchain should follow the XDG base directory standard


mh182:
On a Linux environment the local installation of `docToolchain` is installed in `$HOME/.doctoolchain`.  Linux distributions try to unclutter the user's home directory.

**Describe the solution you'd like**
The local installation of `docToolchain` should follow the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) and install its data into `XDG_DATA_HOME/doctoolchain`. 

An additional benefit would be a cleaner separation between developer environments and installed tool chains.

**Describe alternatives you've considered**
The alternative would be the status quo.

PR #1077, opened 2023-03-09
1031 fix java docker conflict

### LINKED ISSUE #621: dtcw - sdkman installation check returns wrong result
**Describe the bug**
sdkman is installed

**To Reproduce**
I tested this on ubuntu, so here command sdk returns command not found.

**Expected behavior**
sdkman should detected as installed.

**Configuration**
 - OS: Ubuntu 20.10 with zsh

**Additional context**
none

### LINKED ISSUE #973: getJava doesn't work without "local"

### LINKED ISSUE #1031: dtcw ignores installed Java RE when docker is installed - your java version 17 is too new
Following the installation instructions as described in [Install docToolchain](http://doctoolchain.org/docToolchain/v2.0.x/010_manual/20_install.html) with a newer Java version fails.

The reason is due my setup of having a JRE 17 and Docker installed.

I assume users with an installed Docker will prefer the `docToolchain` provided as a container image. But some people may prefer to install the tool chain even if they have docker installed.

**To Reproduce**

System setup: Debian unstable with OpenJDK 17 and docker installed. `docToolchain` was installed with `sdk install doctoolchain`.

```
➜ which java
/usr/bin/java

➜ java -version
openjdk version "17.0.6" 2023-01-17
OpenJDK Runtime Environment (build 17.0.6+10-Debian-1)
OpenJDK 64-Bit Server VM (build 17.0.6+10-Debian-1, mixed mode, sharing)

➜ docker --version
Docker version 20.10.23+dfsg1, build 7155243
```

1. Follow installation instructions as described in [Install docToolchain](http://doctoolchain.org/docToolchain/v2.0.x/010_manual/20_install.html)
```
~/tmp 
➜ mkdir test-doctoolchain && cd test-doctoolchain

~/tmp/test-doctoolchain 
➜ curl -sLo dtcw doctoolchain.github.io/dtcw

~/tmp/test-doctoolchain 
➜ chmod +x dtcw 

~/tmp/test-doctoolchain 
➜ ./dtcw 

dtcw - docToolchain wrapper V0.36
docToolchain V2.1.0
docToolchain as CLI available
docker available
home folder exists
No valid argument(s) supplied
```
2. Due to #1030 I manually changed the wrapper script to use the current tool chain version to v2.2.0
```
~/tmp/test-doctoolchain 
➜ sed -i 's/2.1.0/2.2.0/' dtcw 
```
3. Install Java RE with `getJava`
```
➜ ./dtcw getJava

dtcw - docToolchain wrapper V0.36
docToolchain V2.2.0
docToolchain as CLI available
docker available
home folder exists
this script assumes that you have linux as operating system (x64 / linux)
it now tries to install Java for you
downloading JDK Temurin 11 from Adoptium to /home/max/.doctoolchain/jdk.tar.gz
>>> https://api.adoptium.net/v3/binary/latest/11/ga/linux/x64/jdk/hotspot/normal/eclipse?project=jdk /home/max/.doctoolchain/jdk/jdk.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
100  185M  100  185M    0     0  10.5M      0  0:00:17  0:00:17 --:--:-- 6437k
expanding JDK

~/tmp/test-doctoolchain took 20s
```
3.  Available tasks exists with error since my installed Java version is too new
```
➜ ./dtcw tasks --group=doctoolchain

dtcw - docToolchain wrapper V0.36
docToolchain V2.2.0
docToolchain as CLI available
docker available
home folder exists
use cli install /home/max/.local/share/sdkman/candidates/doctoolchain/current/bin/doctoolchain
Java Version 17
your java version 17 is too new (>14): /usr/bin/java
please update your java installation and try again

it might be that you have installed the needed version java in another shell from which you started dtcw
dtcw is running in bash and uses the PATH to find java

to install a local java for docToolchain, you can run
./dtcw getJava

another way to install or update java is to install
sdkman and then java via sdkman
https://sdkman.io/install
$ curl -s "https://get.sdkman.io" | bash
$ sdk install java

or you can download it from https://adoptium.net/

make sure that your java version is between 8 and 14

If you do not want to use a local java installation, you can also use docToolchain as docker container.
In that case, specify 'docker' as first parameter in your statement.
example: ./dtcw docker generateSite

~/tmp/test-doctoolchain 
❯ echo $?
1
```
4.  Setting `JAVA_HOME` to the previously downloaded JRE is ignored
```
~/tmp/test-doctoolchain 
➜ export JAVA_HOME=/home/max/.doctoolchain/jdk

~/tmp/test-doctoolchain 
➜ $JAVA_HOME/bin/java -version
openjdk version "11.0.18" 2023-01-17
OpenJDK Runtime Environment Temurin-11.0.18+10 (build 11.0.18+10)
OpenJDK 64-Bit Server VM Temurin-11.0.18+10 (build 11.0.18+10, mixed mode)

~/tmp/test-doctoolchain 
➜ ./dtcw tasks --group=doctoolchain

dtcw - docToolchain wrapper V0.36
docToolchain V2.2.0
docToolchain as CLI available
docker available
home folder exists
use cli install /home/max/.local/share/sdkman/candidates/doctoolchain/current/bin/doctoolchain
Java Version 17
your java version 17 is too new (>14): /usr/bin/java
please update your java installation and try again

it might be that you have installed the needed version java in another shell from which you started dtcw
dtcw is running in bash and uses the PATH to find java

to install a local java for docToolchain, you can run
./dtcw getJava

another way to install or update java is to install
sdkman and then java via sdkman
https://sdkman.io/install
$ curl -s "https://get.sdkman.io" | bash
$ sdk install java

or you can download it from https://adoptium.net/

make sure that your java version is between 8 and 14

If you do not want to use a local java installation, you can also use docToolchain as docker container.
In that case, specify 'docker' as first parameter in your statement.
example: ./dtcw docker generateSite
```

**Expected behavior**
After executing `dtcw getJava` the newly installed Java RE is used.

I would change the documentation in [Install docToolchain](http://doctoolchain.org/docToolchain/v2.0.x/010_manual/20_install.html) and make a clear separation between following concepts:
* Installing the tool chain
    * Running the tool chain provided as container image (with `docker` or `podman`)
    * Installing the tool chain to run locally
* Make an existing project (git-repo) ready to use the tool-chain (by adding the wrapper)

Side-note: the output about the missing Java environment is erroneous. Using `sdk install java` will pick the current JRE which is not supported by `docToolchain`.

```
➜ sdk install java

Downloading: java 17.0.6-tem
...
Installing: java 17.0.6-tem
Done installing!

Setting java 17.0.6-tem as default.

allgebras/doctools/docToolchain via ☕ v17.0.6 took 28s 
➜ java -version
openjdk version "17.0.6" 2023-01-17
OpenJDK Runtime Environment Temurin-17.0.6+10 (build 17.0.6+10)
OpenJDK 64-Bit Server VM Temurin-17.0.6+10 (build 17.0.6+10, mixed mode, sharing)

➜ ./dtcw tasks --group=doctoolchain

dtcw - docToolchain wrapper V0.36
docToolchain V2.2.0
docToolchain as CLI available
docker available
use cli install /home/max/.local/share/sdkman/candidates/doctoolchain/current/bin/doctoolchain
Java Version 17
your java version 17 is too new (>14): /home/max/.local/share/sdkman/candidates/java/current/bin/java
please update your java installation and try again
...
```

**Additional context**
Looking into the `dtcw` wrapper script is see that the `JAVA_HOME` is only setup in case docker is not installed.

mh182:
This is a major refactoring of `dtcw` (I would rather say it is a rewrite). So it may make more sense to look at the complete file instead of a diff.

The main goal of the refactoring was to **make the code better understandable**. Please provide feedback if this is true.

Note: the refactoring in `dtcw` is not meant to be complete.

I'm going to focus in the PR on the implementation details. I will explain the user visible changes in #1031.

What changed:
- I replaced the different flags which controlled how `dtcw` behaves with a hash-map. 
- I started to split the code into functions in the hope to make the code better readable.
- Improved test code
  - The tests mock the downloads for fast execution (no internet needed)
  - Split out the end-to-end tests into separate files (`e2e_*.bats`)
  - Improved `dtcw` coverage
  - Use git submodules to install the test framework. This way we can use a standard Debian container image for the tests instead of a home-brewed image.
- I found and fixed various bugs (for example, incorrect handling of JAVA_HOME)

The open questions we should discuss (marked with `# TODO` in `dtcw`):
- I don't fully understand how `DTC_HEADLESS` and `-Dgradle.user.home` work with Docker containers. When do we have to set which variable?
- `DTC_SITETHEME` is never used (the environment flag to the docker container is never passed)
- What is `DTC_PROJECT_BRANCH` good for? (not described anywhere but passed into `dtcw`)
- Why do we check for `.doctoolchain/jdk/Content/Home` when looking for a JRE?
- `generateDeck` can only be called in a `local` environment - is this correct?

What is missing:
- Test coverage (see the `test/README.adoc` file)


mh182:
The Header rules failed. 
I looked into the GitHub workflows but could not find when and how those workflow is executed.
The deployment needs to be adapted.

mh182:
Since I got the `dtcw` tests now working with macOS, if found a major problem with my refactoring. 

macOS uses, due to licensing reasons, an ancient bash version, more precise bash 3.2.

And bash arrays were introduced with bash 4. Since my refactoring uses associative arrays (hash map), it's not going to work on macOS.

mh182:
@rdmueller I have a question: does the current `dtcw` script work on a default bash provided by macOS, or do we expect people updating their bash version with [Homebrew](https://brew.sh/)?

The bash version on macOS 12 is 3.5.57
```bash
$ bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)
Copyright (C) 2007 Free Software Foundation, Inc.
```
I could try to change `dtcw` that it works without index and associative arrays. The code is just going to be hard to maintain. And it doesn't seem that Apple is going to upgrade Bash on their system.

So my question is: should we make Bash >= 4 for `dtcw` a requirement with easier to maintain code or should we not?

mh182:
I found a solution to both problems: 
* GitHub Actions filters `branch-ignore`and `path-ignore` (or `path`) don't work well together.
* Regarding the problem with an old Bash version: I replaced index and associative arrays with strings. Seems to work.

One problem remains: the output and behavior of `dtcw` bash version is now quite different from the MS Windows version. Do we live with that or could you change `dtcw.ps1` based on my changes?

mh182:
@rdmueller Could you please take a look at the changes?

The issues we have to address:
- How can I fix the broken "Header rules" in dtcw-docs-preview step above?
- Decision: do we provide this change as 2.3.0 - `dtcw` has changes in the CLI when installing Java/docToolchain/revealjs. I really would like to get this into production, since it contains bug fixes in corner cases with the different environments.
- Does this version work when we deploy the Docker image (I'm not sure how `gradle.user.home` plays with this)?
- There is a URL in `dtcw`'s output (https://doctoolchain.github.io/docToolchain/#wsl) which leads to nowhere - what do we put in there?
- In general, look at the code marked with `# TODO` maybe you can provide some insights

If we decide this goes into 2.3.0 we have to change the documentation on the page.

rdmueller:
> I don't fully understand how DTC_HEADLESS and -Dgradle.user.home work with Docker containers. When do we have to set which variable?

`DTC_HEADLESS` is used to supress asking if the external theme should be downloaded. This is quite important when on a CI system. 

`-Dgradle.user.home` is the location where dependencies are cached. The docker image provides hopefully all dependencies and the parameter points gradle to those.

> DTC_SITETHEME is never used (the environment flag to the docker container is never passed)

it is used withing `generateSite`. You can specify an external theme which hill be downloaded if it is locally not available. `DTC_HEADLESS` is used to avoid asking if it should be downloaded

> What is DTC_PROJECT_BRANCH good for? (not described anywhere but passed into dtcw)

I think it is mainly used to link to the correct branch for the `improve this doc` link on the right. You can use the branch to create the base link in the config.

* https://github.com/docToolchain/docToolchain/blob/8f8725413b0e2c6fee97656fd0f3a6ca9d2c565a/src/site/templates/rightcolumn.gsp
* https://github.com/docToolchain/docToolchain/blob/a1ef873c2a44cd1d8ebdc506685734cf0832f7f5/docToolchainConfig.groovy

> Why do we check for .doctoolchain/jdk/Content/Home when looking for a JRE?

As far as I remember, the structure of the zipped JDK is different for Windows than for Linux. The windows one contains the `/Content` folder.

> generateDeck can only be called in a local environment - is this correct?

shouldn't be this way but it currently is. Within the docker image, reveal is already cloned into the image:

https://github.com/docToolchain/docker-image/blob/ng-beta/alpine/Dockerfile#L52

But the wrapper ignored it, because something it wrong with the following block:

https://github.com/docToolchain/docToolchain/blob/ng/dtcw#L122

it doesn't recognize reveal as already cloned. Workaround is not to use ´generateDeck` as first parameter. `./dtcw docker install generateDeck` will work. (`docker` is removed as first parameter and `install` is a nearly empty task)

rdmueller:
> The Header rules failed.

afaik, this is a check coming from netlify. 

argh. netlify can't build the website because of the changed dtcw... 

```
12:47:17 PM:   1. Build command from Netlify app                             
12:47:17 PM: ────────────────────────────────────────────────────────────────
12:47:17 PM: ​
12:47:17 PM: $ ./netlifyBuildPreview.sh
12:47:17 PM: dtcw dev - ##DTCW_GIT_HASH##
12:47:17 PM: docToolchain 2.2.1
12:47:17 PM: Available docToolchain environments: local
12:47:17 PM: Environments with docToolchain [2.2.1]: none
12:47:17 PM: Using environment: local
12:47:17 PM: Error: doctoolchain - command not found [environment 'local']
12:47:17 PM: It seems docToolchain 2.2.1 is not installed. dtcw supports the
12:47:17 PM: following docToolchain environments:
12:47:17 PM: 1. 'local': to install docToolchain in [/opt/buildhome/.doctoolchain] use
12:47:17 PM:     $ ./dtcw local install doctoolchain
12:47:17 PM: 2. 'sdk': to install docToolchain with SDKMAN! (https://sdkman.io/)
12:47:17 PM:     # First install SDKMAN!
12:47:17 PM:     $ curl -s "https://get.sdkman.io/" | bash
12:47:17 PM:     # Then open a new shell and install docToolchain with
12:47:17 PM:     $ sdk install doctoolchain 2.2.1
12:47:17 PM: Note that running docToolchain in 'local' or 'sdk' environment needs a
12:47:17 PM: Java runtime (major version 8, 11, or 14) installed on your host.
12:47:17 PM: 3. 'docker': pull the docToolchain image and execute docToolchain in a container environment.
12:47:17 PM:     $ ./dtcw docker tasks --group doctoolchain
```

rdmueller:
> @rdmueller I have a question: does the current dtcw script work on a default bash provided by macOS, or do we expect people updating their bash version with [Homebrew](https://brew.sh/)?

current version seems to work on macOS. Maybe we could try to detect outdated versions and ask the user to upgrade?

rdmueller:
> One problem remains: the output and behavior of dtcw bash version is now quite different from the MS Windows version. Do we live with that or could you change dtcw.ps1 based on my changes?

I think we should align both. I can take a look at it.

dtcw
```diff
@@ -6,84 +6,156 @@ set -e
...
+DTC_ROOT="${HOME}/.doctoolchain"
+
+# More than one docToolchain version may be installed locally.
+# This is the directory for the specific version.
+DTC_HOME="${DTC_ROOT}/docToolchain-${DTC_VERSION}"
+
+# Directory for local Java installation
+DTC_JAVA_HOME="${DTC_ROOT}/jdk"
+
+# The 'generateSite' and 'copyThemes' tasks support DTC_SITETHEME, an URL of a theme.
```
rdmueller: these are the environment vars to configure the wrapper. I think we should distinguish them from the internal script vars and put them on top of the file.

`DTC_VERSION` is the config var which need often to be changed.
`DTC_THEME` needs to be updated once to specify a corporate identity
`DTC_TEMPLATEx` is still experimental, but should be configured once like  `DTC_THEME` to specify non-public templates

dtcw
```diff
@@ -95,311 +167,499 @@ Examples:
...
-            # reveal.js already installed
-            echo ""
-        else
-            echo "cloning reveal.js"
-            (cd "${HOME}/.doctoolchain/docToolchain-${DTC_VERSION}/resources" && ./clone.sh)
-        fi
+  Multiple tasks may be provided at once:
+    ./dtcw generatePDF generateHTML
+
+Detailed documentation how to use docToolchain may be found at https://doctoolchain.github.io/
```
rdmueller: we can switch from doctoolchain.github.io to doctoolchain.org

rdmueller (changes requested):
thanx for the re-write. Seems to me that you are a bash wizard!

I commented on two things I would like to change.
And for the rest I have to admit that I am not too good at bash scripting to fully understand everything. So I would like to give it a try by releasing this as some kind of release candidate.

What do you think?

And btw - we should increase the version to 1.0, shouldn't we?

dtcw
mh182: > these are the environment vars to configure the wrapper. I think we should distinguish them from the internal script vars and put them on top of the file.

They are already on "top of the file". I could move the environment variables which can be provided by the user on the very top.

I don't think it doesn't make sense to distinguish internal and external variables by naming conventions. Common naming conventions for bash "constants" are upper case variables.

Using the prefix `DTC_`would be an indication, but in my experience this changes over time. For example `DTC_ROOT` is internal at the moment, but I would like this to change in the near time future (see #1062).

The information which environment variables are available should be provided by the documentation.

> - `DTC_VERSION` is the config var which need often to be changed. 
> - `DTC_THEME` needs to be updated once to specify a corporate identity
> - `DTC_TEMPLATEx` is still experimental, but should be configured once like `DTC_THEME` to specify non-public templates

Actually, there are more:
- `DTC_OPTS`
- `DTC_PROJECT_BRANCH` 
- `DTC_CONFIG_FILE`

So the only two internal variables left are `DTC_HOME` and `DTC_JAVA_HOME`.  

I could move those two variables down in the file, so users don't see them at the very first glance by skimming over `dtcw`. Would that be fine?

dtcw
mh182: Done

mh182:
> > One problem remains: the output and behavior of dtcw bash version is now quite different from the MS Windows version. Do we live with that or could you change dtcw.ps1 based on my changes?
> 
> I think we should align both. I can take a look at it.

This would be fine. I would not merge this change into ng without the change in `dtcw.ps1`.

We really should think about what technology we should use for v3. Having 3 scripts for the same logic is error prone.

mh182:
> > The Header rules failed.
> 
> afaik, this is a check coming from netlify.
> 
> argh. netlify can't build the website because of the changed dtcw...
> 
> ```

Can we change this per branch basis?

I never worked with netlify, so I have no idea where to adapt this.

mh182:
> > I don't fully understand how DTC_HEADLESS and -Dgradle.user.home work with Docker containers. When do we have to set which variable?
> 
> `DTC_HEADLESS` is used to supress asking if the external theme should be downloaded. This is quite important when on a CI system.

I think this one is set correctly.
 
> `-Dgradle.user.home` is the location where dependencies are cached. The docker image provides hopefully all dependencies and the parameter points gradle to those.

What is the default value if `gradle.user.home` is not provided? Or better, how could I test this?

> > Why do we check for .doctoolchain/jdk/Content/Home when looking for a JRE?
> 
> As far as I remember, the structure of the zipped JDK is different for Windows than for Linux. The windows one contains the `/Content` folder.

OK, I will add this information in the code.

> > generateDeck can only be called in a local environment - is this correct?
> 
> shouldn't be this way but it currently is. Within the docker image, reveal is already cloned into the image:
> 
> https://github.com/docToolchain/docker-image/blob/ng-beta/alpine/Dockerfile#L52
> 
> But the wrapper ignored it, because something it wrong with the following block:
> 
> https://github.com/docToolchain/docToolchain/blob/ng/dtcw#L122
> 
> it doesn't recognize reveal as already cloned. Workaround is not to use ´generateDeck`as first parameter.`./dtcw docker install generateDeck` will work. (`docker`is removed as first parameter and`install` is a nearly empty task)

OK, so the `generateDeck` task should also be available in the other environments. I will have to take a look at this.

mh182:
> thanx for the re-write. Seems to me that you are a bash wizard!
> 
> I commented on two things I would like to change. And for the rest I have to admit that I am not too good at bash scripting to fully understand everything. So I would like to give it a try by releasing this as some kind of release candidate.
> 
> What do you think?
> 
> And btw - we should increase the version to 1.0, shouldn't we?

I would bump the version to 0.50 and go for test period.

But first we have to adjust the changes in `dtcw.ps1`.

mh182:
I adapted the documentation for the installation instructions.

What we are missing now is the changes in the Windows wrapper scripts.

rdmueller:
I currently try to get netlify preview up and running, but I get the following error:

```
4:08:32 PM: + ./dtcw local install doctoolchain
4:08:32 PM: dtcw dev - ##DTCW_GIT_HASH##
4:08:32 PM: docToolchain 2.2.1
4:08:32 PM: Available docToolchain environments: local
4:08:32 PM: Environments with docToolchain [2.2.1]: none
4:08:32 PM: Using environment: local
4:08:32 PM: Installed docToolchain successfully in '/opt/buildhome/.doctoolchain/docToolchain-2.2.1'.
4:08:32 PM: Error: unsupported Java version 8 [/usr/bin/java]
4:08:32 PM: docToolchain supports Java versions 8, 11 (preferred) or 14. In case one of those
```

any idea?

mh182:
> I currently try to get netlify preview up and running, but I get the following error:
> 
> ```
> 4:08:32 PM: + ./dtcw local install doctoolchain
> 4:08:32 PM: dtcw dev - ##DTCW_GIT_HASH##
> 4:08:32 PM: docToolchain 2.2.1
> 4:08:32 PM: Available docToolchain environments: local
> 4:08:32 PM: Environments with docToolchain [2.2.1]: none
> 4:08:32 PM: Using environment: local
> 4:08:32 PM: Installed docToolchain successfully in '/opt/buildhome/.doctoolchain/docToolchain-2.2.1'.
> 4:08:32 PM: Error: unsupported Java version 8 [/usr/bin/java]
> 4:08:32 PM: docToolchain supports Java versions 8, 11 (preferred) or 14. In case one of those
> ```
> 
> any idea?

I will add a test case for Java 8. Could you provide the output of `/usr/bin/java -version` please? Maybe I introduced an error in parsing the output.

I will add a new test case with the output.

rdmueller:
found the problem....

rdmueller:
preview works again: https://deploy-preview-1077--dtc-docs-preview.netlify.app/

rdmueller:
java version is "1.8.0_362", parsing this version works fine - did you use the old script code for parsing it? It already handled some edge cases.

The problem was the comparison:

```
-    if [ "${major_version}" -gt 8 ] && [ "${major_version}" -lt 14 ]; then
+    if [ "${major_version}" -ge 8 ] && [ "${major_version}" -le 14 ]; then
```
`<` vs `<=`

rdmueller:
@mh182 Question:

would it make sense to add some code for backwards compability? something like

[pseudo code]
```
if $1== "getJava"
  $1 = "install java"
  echo "warning: 'getJava' has been replaced by 'install java' and will be deprecated soon
```

?

mh182:
> @mh182 Question:
> 
> would it make sense to add some code for backwards compability? something like
> 
> [pseudo code]
> 
> ```
> if $1== "getJava"
>   $1 = "install java"
>   echo "warning: 'getJava' has been replaced by 'install java' and will be deprecated soon
> ```
> 
> ?

Yes, that would make sense. I will put it on my TODO list.

I just checked the netlify output. All the headers after `Configure docToolchain to Use Existing Documents` are broken. Any clue why? This is already present in the current web-page.

Maybe due to the tags `//end::install[]`?

rdmueller:
> I just checked the netlify output. All the headers after `Configure docToolchain to Use Existing Documents` are broken. Any clue why? This is already present in the current web-page.
> 
> Maybe due to the tags `//end::install[]`?

No, I don't think it is because of this tag. I will check this later.

mh182:
> @mh182 Question:
> 
> would it make sense to add some code for backwards compability? something like
> 
> [pseudo code]
> 
> ```
> if $1== "getJava"
>   $1 = "install java"
>   echo "warning: 'getJava' has been replaced by 'install java' and will be deprecated soon
> ```
> 
> ?

@rdmueller I added the warning and pushed the change.

rdmueller:
what is the meaning of

```
    [ -z "${installations}" ] && installations=" none"
```
?

mh182:
> what is the meaning of
> 
> ```
>     [ -z "${installations}" ] && installations=" none"
> ```
> 

`${installations}` is a string containing all environments in which docToolchain is installed.

The line above means if  docToolchain is not installed (string empty), show `none`. This could be written with

```
if  [ -z "${installations}" ]; then
 Installations=" none"
fi
```

I use boolean logic as oneliner. The part after `&&` is only executed if the first part is true.

rdmueller:
making progress. nearly through

dtcw
```diff
@@ -6,84 +6,157 @@ set -e
...
+    # 'doctoolchain'. This also means the check has to occur at the leatest time.
+
+    command=$(build_command "${environment}" "${DTC_VERSION}" "$@")
+
+    [[ "${DTC_HEADLESS}" = true ]] && echo "Using headless mode since there is no (terminal) interaction possible"
+
+    show_os_related_info
+
+    emu=""
+    if [ "$os" = "Darwin" ] && [ "$arch" = "arm64" ]; then
```
rdmueller: I guess we run into a problem here with `install java`. 
The original idea behind the automatic use of the emulator was not to make the user think, but it will break the build:

The jdk will be installed for the arm architecture and here we switch to x86 architecture.
In addition, the goal is to switch to arm as soon as jbake work on arm.

what do you think?

dtcw
```diff
@@ -6,84 +6,157 @@ set -e
...
+    # The code below assumes we have just one task.
+
+    if [ "${1}" = "generateDeck" ]; then
+        # TODO: is this still true? The Dockerfile at
+        # https://github.com/docToolchain/docker-image/blob/ng-beta/alpine/Dockerfile
+        # seems to install reaveal.js also in the docker image.
+        assert_reveal_js_installed "${environment}"
+    fi
+
+    # If bakePreview is called, deactivate daemon.
```
rdmueller: this can be removed. The daemon is now always deaktivated since we get some memory leaks when using it

rdmueller:
so, the powershell version has been rewritten to match the rewritten bash version.
First manual tests are looking good. Still have to add more automated tests.

What are our next steps?

dtcw
mh182: Well, this is the status quo what the current implementation is doing.
I would like to address the Java/ARM-Mac problem with #1082 and #1092.

I would not release the changes until we addressed a couple other open issues.

dtcw
mh182: I will remove it as soon I'm back from vacation.

mh182:
> What are our next steps?

Check and fix the documentation, address the minor issues from your feedback, manual testing, then integrate it.

Regarding the documentation: have you found the reason why the formatting breaks at the end of the installation section?

I will do that next week after my vacation.

rdmueller:
> Regarding the documentation: have you found the reason why the formatting breaks at the end of the installation section?

I just checked it and couldn't find it anymore?!?

mh182:
Then we have just to update the documentation (output) from the windows ps1 (and bat) scripts.

Since you have a Mac, where do you test the ms windows environment?

rdmueller:
> Since you have a Mac, where do you test the ms windows environment?

Since a few weeks, I own a Mac with M2pro chip for work and I do have a windows machine as my private workhorse. This gives me the flexibility to test everything.

test/install_reveal_js.bats
```diff
@@ -0,0 +1,70 @@
+# SPDX-License-Identifier: MIT
```
PacoVK: Since #1098  this test is now obsolete

test/java_environment.bats
```diff
@@ -0,0 +1,90 @@
...
+    assert_equal "$(mock_get_call_args "${mock_doctoolchain}")" ". tasks --group doctoolchain -PmainConfigFile=docToolchainConfig.groovy --warning-mode=none --no-daemon -Dgradle.user.home=${DTC_ROOT}/.gradle"
+}
+
+@test "invalid JAVA_HOME" {
+    export JAVA_HOME=/jdk/does/not/exist
+    PATH="${minimal_system}" run -1 ./dtcw tasks --group doctoolchain
+    assert_line "Error: unable to locate a Java Runtime"
+}
+
+@test "Java 8 supported" {
```
PacoVK: we should decide on Java 8. Follow discussion on [Gradle 8](https://github.com/docToolchain/docToolchain/issues/1102#issuecomment-1505706457)

dtcw
```diff
@@ -6,84 +6,157 @@ set -e
...
-Use "./dtcw tasks --group doctoolchain" to see available tasks.
-Use "local", "sdk" or "docker" as first argument to force the use of a local, sdkman or docker install.
+    # TODO: can generateDeck, bakePreview be used in combination with other commands?
+    # The code below assumes we have just one task.
+
+    if [ "${1}" = "generateDeck" ]; then
+        # TODO: is this still true? The Dockerfile at
+        # https://github.com/docToolchain/docker-image/blob/ng-beta/alpine/Dockerfile
+        # seems to install reaveal.js also in the docker image.
+        assert_reveal_js_installed "${environment}"
```
PacoVK: this can be removes since #1098

dtcw
```diff
@@ -95,308 +168,509 @@ Examples:
...
+    # TODO: should we not delete a previsouly installed on?
+    # Otherwise we may get a mix of different Java versions.
+    tar -zxf "${DTC_JAVA_HOME}/jdk.tar.gz" --strip 1 -C "${DTC_JAVA_HOME}/."
+    rm -f "${DTC_JAVA_HOME}/jdk/jdk.tar.gz"
+
+    echo "Successfully installed Java in '${DTC_JAVA_HOME}'."
+    echo
+}
+
+local_install_reveal_js() {
```
PacoVK: can be removed since #1098

dtcw
```diff
@@ -95,308 +168,509 @@ Examples:
...
+        "Note that running docToolchain in 'local' or 'sdk' environment needs a" \
+        "Java runtime (major version 8, 11, or 14) installed on your host." \
+        "" \
+        "3. 'docker': pull the docToolchain image and execute docToolchain in a container environment." \
+        "" \
+        "    \$ ./dtcw docker tasks --group doctoolchain" \
+        ""
+}
+
+assert_reveal_js_installed() {
```
PacoVK: can be removed since #1098

dtcw.ps1
```diff
@@ -255,24 +357,211 @@ else {
...
+    Write-Host "Extracting JDK from archive file."
+    # TODO: should we not delete a previsouly installed on?
+    # Otherwise we may get a mix of different Java versions.
+    Expand-Archive -Force -LiteralPath "$DTC_JAVA_HOME\jdk.zip" -DestinationPath "$DTC_JAVA_HOME"
+    Remove-Item "$DTC_JAVA_HOME\jdk.zip"
+
+    Write-Host "Successfully installed Java in '$DTC_JAVA_HOME'."
+}
+
+function local_install_reveal_js() {
```
PacoVK: can be removed since #1098

test/e2e_local_install.bats
```diff
@@ -0,0 +1,94 @@
...
+@test "show available tasks" {
+    run -0 ./dtcw tasks --group doctoolchain
+
+    # TODO: 'gradlew' is not visible for the usevisible for the user
+    assert_line "To see all tasks and more detail, run gradlew tasks --all"
+    assert_line "To see more detail about a task, run gradlew help --task <task>"
+}
+
+# bats test_tags=e2e
+@test "install reveal.js" {
```
PacoVK: can be removed since #1098

test/install_reveal_js.bats
```diff
@@ -0,0 +1,70 @@
...
+    # Execute
+    PATH="${minimal_system}" run -1 ./dtcw install reveal.js
+
+    # Show what the user should do
+    assert_line "Error: git - command not found"
+
+    assert_line  "Installing reveal.js depends on 'git' - please install 'git' (https://git-scm.com)."
+}
+
+@test "install reveal.js" {
```
PacoVK: can be removed since #1098

PacoVK (commented):
Great rewrite! I skimmed through the change and added some comments that may ease the rebase on `ng`

mh182:
I rebased this PR and removed the code to install reveal.js.

rdmueller (approved):

mh182:
I really would like to get this PR into the `ng` to avoid continuous rebasing.

@rdmueller Is the  re-write of `dtcw.ps1` complete?

rdmueller:
@mh182 , yes, the rewrite should be complete. If we.merge it now, I also have time to fix upcoming problems quickly

dtcw
mh182: I will address this topic in #1082.

dtcw
```diff
@@ -95,300 +152,476 @@ Examples:
...
+show_os_related_info() {
+
+    # check if we are running on WSL
+    if grep -qsi 'microsoft' /proc/version &> /dev/null ; then
+        # TODO: bug - This URL leads to now-where!
+        printf "%s\n" \
+            "" \
+            "Bash is running on WSL which might cause problems with plantUML." \
+            "See https://doctoolchain.github.io/docToolchain/#wsl for more details." \
+            ""
```
mh182: @rdmueller This link leads to the docToolchain landing page (is not valid anymore). I could not find information about WSL with plantuml anymore.

src/docs/010_manual/20_install.adoc
```diff
@@ -146,71 +246,74 @@ To get started with docToolchain, run `tasks` as the first command:
...
+DTC_OPTS="-PmavenRepository=your_maven_repo -PmavenUsername=your_username -PmavenPassword=your_pw" ./dtcw tasks --group=doctoolchain --info
 ----
 
-You'll notice the installation is an incremental process and will take a while because of all the additional dependencies needed.
-After the first few commands, you'll have all the required dependencies installed and things will run faster.
-
 //end::install[]
 
-== Configure docToolChain to Use Existing Documents
+== Configure docToolchain to Use Existing Documents
```
mh182, line 278: @rdmueller Lookig at the netlify preview sections after here are not rendered correctly. No clue why.

changelog.adoc
```diff
@@ -9,16 +9,24 @@ and this project tries to adhere to https://semver.org/spec/v2.0.0.html[Semantic
...
-* fix decision tree for `dtcw.ps1`
-* #831 : fix for generateSite: images-hrefs not correct
-* #1109 : docToolchain release notes contain releases twice
+* https://github.com/docToolchain/docToolchain/issues/621[#621: dtcw - sdkman installation check returns wrong result]
+* https://github.com/docToolchain/docToolchain/issues/831[#831: fix for generateSite: images-hrefs not correct]
+* https://github.com/docToolchain/docToolchain/issues/973[#973: `dtcw getJava` doesn't work without `local`]
+* https://github.com/docToolchain/docToolchain/issues/1109[#1109: docToolchain release notes contain releases twice]
+* https://github.com/docToolchain/docToolchain/issues/1031[#1031: dtcw ignores installed Java RE when docker is installed - your java version 17 is too new]
+* various fixes in `dtcw`, `dtcw.ps1`:
+** fix decision tree for `dtcw.ps1`
```
mh182: @rdmueller What decision tree did you mean here? I can't understand what you mean.

src/docs/010_manual/20_install.adoc
```diff
@@ -101,42 +89,154 @@ curl -Lo dtcw.bat doctoolchain.github.io/dtcw.bat
...
+<.> List of available docToolchain environments. The output may vary depending on your system. In our example only the `local` environment is available since neither `sdk` nor `docker` was found.
+<.> Environments in which docToolchain s available. The output may vary depending on how you installed docToolchain. In our example docToolchain was found in the user's local environment in `$HOME/.doctoolchain`.
+<.> Shows the used docToolchain environment. In case docToolchain is installed in more than one environment the wrapper script picks the environment in the following order: `local`, `sdk`, and then `docker`.
+<.> Location of the used Java. In our example Java was installed in the local environment with the docToolchain wrapper script.
+<.> docToolchain was invoked the first time, thus it is downloading its dependencies.
+<.> The docToolchain configuration file `docToolchainConfig.groovy` wasn't found in the project repository. docToolchain asks if it should create a new one.
+
 --
 
 .Windows with Powershell
```
mh182, line 242: How does the new output of `dtcw.ps1` look like?

mh182 (commented):
Minor issues to fix Last fixes before we can integrate.

dtcw
rdmueller: correct link is https://doctoolchain.org/docToolchain/v2.0.x/010_manual/50_Frequently_asked_Questions.html#wsl
I'll fix this

rdmueller: fixed

src/docs/010_manual/20_install.adoc
rdmueller: found it. fixed it.
Seems that the `--` blocks where nested wrong

changelog.adoc
rdmueller: the decision which installed version to use. docker, local, sdkman and if java has to be installed

mh182: Ahh, OK. I try to describe this in more details.

src/docs/010_manual/20_install.adoc
rdmueller: good point. let me fix this...

mh182: OK, updating the documentation is the only thing missing, ATM.
Would be nice if we could close this issue since it keeps other PRs blocked.

rdmueller: I currently extend the dtcw.ps1 a little bit in order to make it work with linux environments.
That lets us run tests on our linux based build servers and lets us capture the output for the documentation. I think I can finish this change today. 

Let's merge these thenges afterwards and move the docs update to another issue?

rdmueller: ok. I checked the install docs file and updated what was needed to be updated. Noticed that there is no powershell output :-)

rdmueller (approved):
I am quite happy with these changes. let's merge!

dtcw
```diff
@@ -163,7 +165,7 @@ EOF
 }
 
 print_version_info() {
-    echo "dtcw ${DTCW_VERSION} - ${DTCW_GIT_HASH}"
+    echo "dtcw ${DTCW_VERSION} (${os} ${arch}) - ${DTCW_GIT_HASH}"
```
mh182: I'm not really happy about this change.
The `--version` output is also meant to be easy parsable by scripts (for example in the CI build).
Parsing for a version number on `dtcw 0.50 (Linux x86_64) - ##DTCW_GIT_HASH##` doesn't make is easier.
What about we change the output for the OS/architecture in a separate line?
```
dtcw 0.50 - ##DTCW_GIT_HASH##
docToolchain 2.2.1
OS/Arch: Linux/x86_64
```

mh182: I will change this on all the scripts.

mh182: And I was already wondering why the tests didn't fail. 
The reason was I dynamically parse the version from `--version` which now had the value `0.5.0 (Linux x84_64)`.

rdmueller: OK. I did this change because a simple `./dtcw` command didn't output the version.
And I wanted to see on which system the script is running. I think this is important when users file issues.
Maybe we can modify it in such a way that all requirements are met?

mh182: I specifically didn't show version information if someone just started the script without any argument. Maybe not so a good idea.
I'm going to rebase, squash everything and merge.

PR #1078, opened 2023-03-10
dtcw uses main_config_file environment variable

### LINKED ISSUE #995: Wrapper uses external main_config_file variable
Dear all, 
in the current implementation of the dtcw / dtcw.ps1 we do not see a solution for specifying a custom main configuration file with name and location other than docToolchainConfig.groovy in the root folder of a project.

So we have to adjust the hardcoded line 
`MAIN_CONFIG_FILE=docToolchainConfig.groovy` (dtcw line 15) 
to something like 
`MAIN_CONFIG_FILE=doc/src/docs/docToolchainConfig.groovy`

I would recommend to replace the current logic to something like ...
 
```
if [ "$MAIN_CONFIG_FILE" = "" ] ; then
  MAIN_CONFIG_FILE=docToolchainConfig.groovy
fi
```

... or do you see any other solution ?

Can also provide a PR for that.

RoettingerJ:
resolves #995 

### All Submissions:

### Your first submission


dtcw
```diff
@@ -14,7 +14,11 @@ set -o pipefail
 #export DTC_TEMPLATE2=https://....zip
 #...
 
-MAIN_CONFIG_FILE=docToolchainConfig.groovy
+if [[ -z "${MAIN_CONFIG_FILE:-""}" ]]; then
+  DTC_MAIN_CONFIG_FILE=docToolchainConfig.groovy
+else
+  DTC_MAIN_CONFIG_FILE="${MAIN_CONFIG_FILE}"
+fi
```
mh182: Since the environment variables are used by the user to tweak the behavior of docToolchain we should standardize the naming conventions for the environment variables which are exposed to the user.

The other variables use the `DTC_` prefix. What about `DTC_CONFIG_FILE`?

I don't like the term "main config file" since it suggests the existence of additional configuration files. So far, I haven't seen other user exposed configuration files.

If we use `DTC_CONFIG_FILE` we should change the Gradle parameter to `-PconfigFile` too (which doesn't have to be done in this PR).

I see no point to use two variables. 

The code would look like
```sh
: ${DTC_CONFIG_FILE:-docToolchainConfig.groovy}

mh182 (changes requested):
We should ask @rdmueller if he is fine to use the `DTC_CONFIG_FILE`. In case we get a yes, we decide we change the variable name.

dtcw
rdmueller: looks good to me. And yes, I guess it would be good practice to prefix all our environment variables with "DTC" = _d_oc_T_ool_c_hain

mh182: If we use `DTC_CONFIG_FILE` do we change the Gradle parameter from `mainConfigFile` to `configFile`?

This way developers don't have to make a mental mapping, and it makes the code cleaner.
Since it is an internal change, we could do that in a separate issue.

rdmueller: This will be a breaking change, won't it?
Yes, we should change it but I wonder if we should also support `mainConfigFile` for backwards compability without documenting it. What do you think?

mh182: This really depends how you define the interface. Since the whole point of this issue was that the config name was not configurable it was not exposed via doctoolchain nor via dtcw.

I don't know how grade works if someone would provide the `manConfigFile` a secon time vie `DTC_OPT`.

So I see it more as this parameter was newer exposed.

But if it doesn't make much work, we could add the new parameter `configFile` which maybe overruled by the existing on and provide a warning.

But we can discuss this in issue we are going to make.

rdmueller (approved):
fine for me with the described changes

mh182:
I thought we agreed to call it `DTC_CONFIG_FILE` to not confuse users about possible other configuration files?
Be aware that I don't mean to change the Gradle parameter `mainConfigFile`.

dtcw
rdmueller: ok. I guess you are right, we just expose it as new parameter, so let's do it!

rdmueller:
> I thought we agreed to call it `DTC_CONFIG_FILE` to not confuse users about possible other configuration files? Be aware that I don't mean to change the Gradle parameter `mainConfigFile`.

Was my review comment confusing? I just wanted to close this review. Yes, it should be called `DTC_CONFIG_FILE` and we need the rquested change for that :-) Who will do it? @RoettingerJ ?

RoettingerJ:
Hopefully understood everything right ;-) ...

mh182 (approved):

mh182:
@RoettingerJ I adapted the description in the changelog. Otherwise it looks fine.

Could you please rebase your implementation to the current `ng` branch so we can integrate it?

Thanks.

RoettingerJ:
Hm should be already up to date - or what do you expect exactly ?

mh182:
Sorry my fault. It was already rebased, and thus the GitHub UI disabled the "rebase" button. I'm a noob with GitHub UI workflows.

mh182:
Thanks for your contribution.

PR #1084, opened 2023-03-16
Overwrite arch with x86_64 in case of i386

### LINKED ISSUE #1082: Not able to use dtcw on arm64 / apple silicone
**Describe the bug**
I'm not able to run the dtcw wrapper locally on my Mac. 

_For running the local version when running `donwloadTemplate`_:
```

* Where:
Settings file '/Users/simon.weiss/.doctoolchain/docToolchain-2.2.1/settings.gradle'

* What went wrong:
Could not compile settings file '/Users/simon.weiss/.doctoolchain/docToolchain-2.2.1/settings.gradle'.
> startup failed:
  General error during semantic analysis: Unsupported class file major version 61

  java.lang.IllegalArgumentException: Unsupported class file major version 61
```
_For running `./dtcw local generateSite`_ I get the error that is probably connected to orient-db https://github.com/jbake-org/jbake/issues/709:
```

* What went wrong:
Execution failed for task ':bake'.
> A failure occurred while executing org.jbake.gradle.impl.JBakeWorkAction
   > Can't load library: /var/folders/9l/s2lzkzbn2y98d96lgr8q961r0000gn/T/jna-1355430915/jna10246597454490203081.tmp
```
 
_For running via docker_, e.g. `./dtcw docker generateSite` it just hangs and never responds. This might be because it needs to emulate the x86_64 architecture.

**To Reproduce**
1. Download dtcw
2. Run `./dtcw downloadTemplate` or `./dtcw docker generateSite`on Mac M1 

I also tried it with the `arch -x86_64 /bin/bash` like outlined in the help message, however, this didn't help. 

**Expected behavior**
Generating output files, downloading template, etc

**Configuration**
 - docToolchain version 2.2.1
 - OS: MacOS
 - Relevant SDK Versions: 
 ```
openjdk 11.0.17 2022-10-18
OpenJDK Runtime Environment Temurin-11.0.17+8 (build 11.0.17+8)
OpenJDK 64-Bit Server VM Temurin-11.0.17+8 (build 11.0.17+8, mixed mode)
```

**Additional context**
Relevant issue outlining the problem with orient-db in jbake repo: https://github.com/jbake-org/jbake/issues/709.
Proposed solution is to update jbake: https://github.com/jbake-org/jbake/issues/709#issuecomment-1407811306

Linking other, similar issues to avoid too many duplicated issues: https://github.com/docToolchain/docToolchain/issues/1015
Docker related (not providing an arm image): https://github.com/docToolchain/docker-image/issues/26

### PR DESCRIPTION
**Pull request by SimonCW**
### All Submissions:

### Your first submission

As discussed here: https://github.com/docToolchain/docToolchain/issues/1082

### CODE REVIEW

**Comment by rdmueller** [comment]
on a windows machine, `arch` will return `x86_64`, so this mod will be fine

**Comment by rdmueller** [comment]
@mh182 please merge if you also think this is ok

**Comment by mh182** [comment]
To be honest, just looking at those 5 lines of codes, in a couple of weeks, no one is going to understand them. At least I would not.

`man arch` gives me `arch - print machine hardware name (same as uname -m)`.

So if we don't provide some context with the code (and maybe a reference to ticket #1082) I would refactor the code and delete those lines  in a couple of weeks.

@SimonCW So how about providing a bit of documentation and explain why we have this strange code in there?

File: dtcw [hunk 1]
```diff
@@ -54,10 +54,14 @@ sdkman=false
 homefolder=false
 java=false
 doJavaCheck=true
-arch=$(uname -m)
+if [ "$(arch)" == "i386" ]; then
+  arch="x86_64"
+else 
+  arch=$(uname -m)
+fi
```
mh182, line 61: Please provide some context within the code why we do this.
Additonal a reference to the ticket where we discussed the problem may help.

mh182 (changes requested):

SimonCW:
TBH, I'm not sure whether it should be in there at all. It might be that it is just a weirdness of my config on my Mac that it outputs:
```
$ arch -x86_64 /bin/zsh
$ arch
i386
$ uname -m
arm64
```
I didn't have time to investigate on my machine or research whether this is expected behavior. Just wanted to share my solution. I can check with a few colleagues that have the same machine next week. And try to find out how reliable arch / uname is for determining architecture.



😜

dtcw
rdmueller: We had some unexpected behaviour of the `uname` command. I can't remember - it returned "" or "arm64" but the system was in emulated i386 mode. The `arch` command provided better results, but it returns other identifiers.
So, this code helped to solve this issue.

What do you think about it?

rdmueller:
So, regarding the answer of chatgpt, `uname -m` is the wrong thing to do and it should always reflect the hardware architecture not the one of the emulator?

dtcw
mh182: Damn, Apple doesn't make it easy. Searching the web, I found [a comment on how to detect Apple Silicone on command line](https://stackoverflow.com/a/68148776).

Nevertheless, I don't think this is the right place to make the mapping. We map the `arch` environment variable at https://github.com/docToolchain/docToolchain/blob/6178309da8968c8f99ba1575faacce885677575f/dtcw#L292-L295 again with the goal download the correct JVM.

IMHO it doesn't make sense to map the output of `uname -m` (or `arch`) twice. I would like to provide a clean solution with #1092. 

Does that make sense?

raifdmueller:
Hi, and thank you for this pull request! 🙏

Let me be honest: in the past we simply didn't have the time for someone to give the open PRs the attention they deserve — and that's why this one has been sitting here untouched. We're sorry about that, because every PR represents real work and good intentions.

We're now trying to work through this backlog with the help of Claude Code. As part of that, we'll **go through each PR individually and decide whether it's still relevant or not**. Please bear with us if a PR ends up being closed — that's not a judgement on your work, it's just part of the cleanup.

**Important note:** We're currently working on **docToolchain v4**: https://doctoolchain.org/docToolchain/v4.0.x/ — so some changes may already be outdated or may need to be integrated differently.

If you'd like to support the project (time is our scarcest resource 😉), we'd really appreciate it via either of these links:
- https://opencollective.com/doctoolchain
- https://github.com/sponsors/rdmueller

Thanks again for contributing, and for your patience!

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1104, opened 2023-04-11
Upgrade/dependencies

PacoVK:
## What this PR does
There are some outdated dependencies and also an older Gradle version. Due to constraints with the current ascciidoctor gradle plugin version 3.x we cannot upgrade Gradle to > 7.0.2. [see](https://github.com/asciidoctor/asciidoctor-gradle-plugin/issues/626). Some other dependencies also lack support for Gradle >= 8.x.

**TL;DR;** upgrade all dependencies and Gradle to the maximum latest version that is currently not version alpha and not breaking tasks due to version constraints

### All Submissions:


mh182:
Once again, your work is appreciated. 

Since this PR touches a lot of tasks my question is how could we assure that we don't break existing functionalities? Do we have tests for the tasks which are affected by this PR?

I have no clue how we could test Jira and Confluence tasks since I have no access to neither of those platforms.

I would like to wait until @rdmueller is back from vacation before we merge this PR. I don't have the knowledge how I could test/verify, if we break something with this PR.

PacoVK:
Totally agree with you! Most of the Tasks have tests, and fortunately some of them broke, which is a good sign 🤪 
In addition i did some manual testing, maybe i can make it to test against a real confluence. 

For that it may be usefull to sign up for the [Atlassian Open Source Programm](https://www.atlassian.com/software/views/open-source-license-request) I think docToolchain fullfills all requirements there.

rdmueller:
a) We already have an "open source program" confluence: https://arc42-template.atlassian.net/wiki/spaces I can give you access when I find the time

b) I also did some work on the dependencies. It is not only that we have the latest dependencies but I also noticed that (now that have some security scans) we could exclude some not used libraries to close some security findings.

PacoVK:
Cool, let's see if we can join forces on the dependencies then. With Gradle 8 we could use dependency version catalog for a even better management of those

mh182:
I just checked the docToolchain home page created by this PR. 

The only difference I found between the current home page at https://doctoolchain.org/docToolchain/v2.0.x/index.html and https://deploy-preview-1104--dtc-docs-preview.netlify.app/ is the missing icon on the front-page (which was added by @rdmueller last week).

PacoVK:
not sure about the image if it's really related to the PR here. I am a bit confused and seems to be broken already in #1096 [see](https://deploy-preview-1096--dtc-docs-preview.netlify.app/)which was directly on top of the commits that added the image. Basically i found no preview link that displayed the image correctly

mh182 (approved):

rdmueller:
let's not worry about the image. If it is really broken, I can fix it after the merge.

Issue #1105, opened 2023-04-12
Upgrade Asciidoctor Gradle Plugin v4


mh182:
Created due to the discussion in #1102.

docToolchain uses Asciidoctor Gradle Plugin v3.*, which is in maintenance mode only. We can not upgrade to a newer Gradle version which supports Java 17 due to [asciidoctor-gradle-plugin/issues/626](https://github.com/asciidoctor/asciidoctor-gradle-plugin/issues/626). 

With this issue, we want to test the maturity of Asciidoctor Gradle Plugin v4.

PacoVK:
see #1108 

@mh182  sorry, messed it a bit up i was not aware that i could not create a PR for another open PR, hence #1108 also contains all changes of #1104  

Do you think we should do the evaluation based on Gradle 6.x (current ng-branch) if possible? Not sure what is best here

mh182:
> @mh182 sorry, messed it a bit up i was not aware that i could not create a PR for another open PR, hence #1108 also contains all changes of #1104

Thanks a lot. 

To make life easier for integrating the code, could you squash all commits of #1104 into a single commit?
It would not only make #1108 easier to integrate, but also make it easier for me to rebase #1077 on top of #1104.

> Do you think we should do the evaluation based on Gradle 6.x (current ng-branch) if possible? Not sure what is best here

I have no idea how @rdmueller tested the correctness of all the different features of docToolchain. That's we have to discuss when he is back from vacation.

Is there a reason you upgraded to Gradle 7.5.1 and not the latest 7 version, which would be 7.6.1?

PacoVK:
Sure I will do that.

Concerning Gradle version unfortunately there is an incompatibility from 7.6 onwards with the Asciidoctor Gradle plugin see [issue](https://github.com/asciidoctor/asciidoctor-gradle-plugin/issues/655) I would not suggest to try some of the mentioned workarounds but rather wait for the Gradle 8 support and just skip 7.6 then

mh182:
OK, good to know.

This means - assuming the result with the plugin are OK - we could upgrade to Gradle 7.5, which supports Java 17, the current LTS. 

This way we would have time until September 2023 when [Java 21, the next LTS](https://en.wikipedia.org/wiki/Java_version_history) is released.

PacoVK:
That's a plan 😊 I check for Java 17 if there are issues and will then enable it in the CI

mh182:
[...]

PR #1108, opened 2023-04-12
Test maturity of Asciidoctor Gradle Plugin v4

### LINKED ISSUE #662: Upgrade to Gradle (wrapper) 7.5
Gradle is only compatible to latest LTS JDK (17) with version 7.x. Therefore I suggest to upgrade to Gradle (wrapper) version 7.2.

PacoVK:
Relates to #1105  this is a PoC for the ascidoctor gradle plugin suite v4.x which is in alpha state as of now. v3.x is in maintenance mode and blocks us from upgrading to Gradle > 7.0.2 . This PR also upgrades Gradle to 7.5.1 the maximum supported version as of now. 

### All Submissions:

### DIFF
--- a/.github/workflows/default-build.yml
+++ b/.github/workflows/default-build.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         # TODO: Will it really run on one of these? , windows-latest, macOS-latest]
-        java: [8, 11, 14] # , 17]
+        java: [8, 11, 14, 17]
         jdk: [adopt]
         # , temurin]
       fail-fast: false
--- a/changelog.adoc
+++ b/changelog.adoc
@@ -27,7 +27,7 @@ and this project tries to adhere to https://semver.org/spec/v2.0.0.html[Semantic
 ** removal of cloning `reveal.js` (`clone.sh` and `clone.ps1` scripts?
 * Doctoolchain now fails if there are images that cannot be resolved during build
 * Upgrade Gradle to 7.5.1
-* Introduce Java 17 support
+* Add support for Java 17
 * Upgrade dependencies
 ** 'org.apache.httpcomponents:httpmime:4.5.14'
 ** 'org.codehaus.groovy:groovy-xml:3.0.13'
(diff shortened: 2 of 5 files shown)


mh182:
The netlify page looks good (apart from the missing icon, which is already present in #1104).

mh182:
@rdmueller could you decide if we are going to integrate the Gradle plugin v4 which would open the migration path to Gradle 7.5 and thus Java 17 (after we integrate #1104).

As mentioned before, I don't have the expertise on the full functionality of docToolchain. 

Apart from infrastructure upgrade, the PR would also make the installation of revealjs with `dtcw` obsolete.

rdmueller:
@mh182 the change will only be in "production" as soon as we create a release with an incremented version number in dtcw.

if this breaks some functionality, people can just switch back the version number in their dtcw.
Existin project will still use the old version.

What if we create a new version "vx.y.z-RC1" and don't increment the version in dtcw? This would allow all users to test the releasecandidate by manually setting the version number.

Another option is to find some time to execute more manual tests for a wider range of tasks.

rdmueller (approved):
I checked all changes - mostly version changes. These look good.
I didn't test the code.

mh182:
> @mh182 the change will only be in "production" as soon as we create a release with an incremented version number in dtcw.
> 
> if this breaks some functionality, people can just switch back the version number in their dtcw. Existin project will still use the old version.
> 
> What if we create a new version "vx.y.z-RC1" and don't increment the version in dtcw? This would allow all users to test the releasecandidate by manually setting the version number.
> 
> Another option is to find some time to execute more manual tests for a wider range of tasks.

I did not ask to release the current functionality (which we should discuss sooner than later). I'm more concerned that the upgrade of those dependencies may break some functionalities.

My main problem is, that I can't test the functionalities provided by docToolchain. So I can not do any QA. This is one of my major hindrances to integrate changes without your help.

So I will merge the upgrade of the plugin v4 and the support for Java 17.

mh182:
@PacoVK I merged #1104. Could you rebase this PR and fix the conflicts so I can merge it into ng?

mh182:
I integrated the PR after some minor changes.

mh182:
Reopen the PR. The CI build broke with my changes.

mh182:
Closing again. Upgrade to Gradle plugin v4.* done.

I had to disable tests for Java 17 in the CI build since one test installed the current docToolchain-2.2.1 which fails since it uses Gradle 6.9.1. I have no idea why your tests passed.

PacoVK:
> Closing again. Upgrade to Gradle plugin v4.* done.
> 
> I had to disable tests for Java 17 in the CI build since one test installed the current docToolchain-2.2.1 which fails since it uses Gradle 6.9.1. I have no idea why your tests passed.

Strange i can check if i find something out there and may provide a fix then

PR #1126, opened 2023-04-26
Fix/1124 set correct outdir for pdf

### LINKED ISSUE #1124: generatePDF: wrong target directory
when `generatePDF` is executed on the arc42 template, the pdf is written to `build/arc42/arc42.pdf` but it is expected in `build/pdf/arc42/arc42.pdf`

PacoVK:
Closes #1124 configure the correct output dir for PDF task


PacoVK:
duplicate of #1125  hence we can close this

PR #1130, opened 2023-04-28
dtcw: drop support for Java 8

### LINKED ISSUE #1116: Drop support for Java 8
With #1071 we support Java 17, the current Java LTS.

Java 8 reached its EOL on [March 2014](https://www.oracle.com/java/technologies/java-se-support-roadmap.html). More and more dependencies which docToolchain is based on do not work with Java 8 anymore.

Note: we have to make sure that Java 8 can be used with docToolchain 2.2.1 and prior since this is supported by the tool chain.

mh182:


mh182:
@rdmueller Hmm, the Netlify deployment uses Java 8. Where do we change the Netlify configuration?

rdmueller:
We xould do a
dtcw install java
Here https://github.com/docToolchain/docToolchain/blob/ng/netlifyBuildPreview.sh

PacoVK:
Seem like Java 17 is working properly. I could not reproduce any failure (locally and on CI)

PacoVK (approved):
LGTM, thanks :)

PR #1135, opened 2023-04-28
add spock tests for powershell to github action

### LINKED ISSUE #1134: CI build: automated tests for dtcw.ps1
Improve CI build introduced by #1073 for `dtcw.ps1`.

The existing tests should be executed on changes on
- `dtcws.ps1`
- the test code testing the script

I think it would make sense to run the CI build on a GHA MS Windows runner.

Note: we don't have the same test coverage as with `dtcw`. The goal is to improve the CI build.

rdmueller:


rdmueller:
"All checks have passed" - we are ready to merge.

mh182 (changes requested):
Would it not make sense to move the tests for `dtcw.ps` from `src/test/groovy/docToolchain` to `test/dtcw.ps1`? It probably would make sense to move the tests for the bash script into `test/dtcw`.
The wrapper scripts semantically do not belong to docToolchain itself. That's why I started the tests for the wrapper scripts somewhere else.

When we implement #1149 this separation will be more clear.

mh182:
Could you squash the commits into one?

rdmueller:
@mh182 commits can be squashed when merged

regarding the location of the tests: `src/test/groovy` is the standard for spock tests and I guess we would even have to reconfigure spock if we move them outside. I would like to post pone this refactoring.

regarding if it belongs to docToolchain or not, I still consider the wrapper to be a core part of docToolchain :-)

mh182:
> @mh182 commits can be squashed when merged

Does the squash merge also rebase? 

> regarding the location of the tests: `src/test/groovy` is the standard for spock tests and I guess we would even have to reconfigure spock if we move them outside. I would like to post pone this refactoring.

I understand. So let's keep it where it is at the moment.

> 
> regarding if it belongs to docToolchain or not, I still consider the wrapper to be a core part of docToolchain :-)

I agree with you, but the wrapper is optional (a side car). "Design for deletion" is a good concept. Just imagine that we want to get rid of the wrapper code. How hard is it for someone to delete the code which is related to the wrapper scripts?

So from my parts lets merge.

mh182 (approved):

rdmueller:
I guess a squash implies a rebase. Lets give it a try

Issue #350, opened 2019-08-26
Autoconfig for needed conversions


rdmueller:
The task `convertToDocx` depends on `generateDocbook`.
So, it would be great if one would only need to configure `docx` in the docToolchain configuration and the `generateDocbook` task would also be - automatically - configured.

In addition, it would be great if these configs are case insensitive.

Issue #433, opened 2020-06-08
Dockercontainer & Pandoc


dr460neye:
**Describe the bug**
Seems like pandoc is not part of the docker container atm.

**To Reproduce**
Search for the pandoc inside the newest container release. Try to run adoc to pandoc conversion.
Both seems to fail.

**Expected behavior**
Container is able to convert adoc to docx

**Configuration**
 - Docker

```

outputPath = 'build/docs'

// Path where the docToolchain will search for the input files.
// This path is appended to the docDir property specified in gradle.properties
// or in the command line, and therefore must be relative to it.

inputPath = 'Folder/';
inputFiles = [
        [file: 'file.adoc', formats: ['html','adoc','docbook','docx']],
]
taskInputsDirs = ["${inputPath}/media"]
taskInputsFiles = []
```

`doctoolchain . --stacktrace --info convertToDocx -PmainConfigFile=....`

ehmkah:
I opened a PR for this requirement. I am happy to hear your feedback

ehmkah:
Now there is a docker-image with pandoc available - https://hub.docker.com/r/doctoolchain/doctoolchain-with-pandoc

I believe this issue could be closed.

PR #1137, opened 2023-04-29
Improve failure handling for transitive task configuration

### LINKED ISSUE #220: convertToDocx and convertToEpub not working
## Expected behavior and actual behavior.
docToolchain/bin/doctoolchain . convertToDocx
The above command works until after the commit 552a12371a09efd79eac111a6869aefb6d38a01d on scripts/pandoc.gradle

I made a small change onto the file under task convertToDocx as follow: 
-def sourceFilesDocx = sourceFiles.findAll { 'docx' in it.formats }
+def sourceFilesDocx = sourceFiles.findAll { 'docbook' in it.formats }

It works again. Please confirm and fix the master file if it is correct.

However, there is another similar issue on running convertToEpub, which is related to the same file but it isn't able to fix by changing docx to docbook under convertToEpub task

-def sourceFilesEpub = sourceFiles.findAll { 'epub' in it.formats }
+def sourceFilesEpub = sourceFiles.findAll { 'dockbook' in it.formats }

Always get the below error
Execution failed for task ':convertToEpub'.
> execCommand == null!

## Steps to reproduce the problem.
checkout out docToolchain project
run ./gradlew -b init.gradle initArc42EN -PnewDocDir=/tmp/Arc42EN
cd /tmp/Arc42EN
${docToolchain_home}/bin/doctoolchain . convertToDocx

## Specifications like the version of the project, operating system, or hardware.
The latest version
I am on fedora 28 but it should be the same problem on other OS platforms

PacoVK:
Closes #220 

print warning instead of throwing an exception when config for input files is missing. Default will convert all files registered as input

### All Submissions:

### DIFF
--- a/src/test/groovy/docToolchain/GenerateDeckSpec.groovy
+++ b/src/test/groovy/docToolchain/GenerateDeckSpec.groovy
@@ -32,9 +32,8 @@ class GenerateDeckSpec extends Specification {
                     .withArguments(['generateDeck','--info', '-PinputPath=src/test/docs','-PmainConfigFile=src/test/config_without_revealjs.groovy'])
                     .build()
         then: 'we get an exception'
-            def e = thrown java.lang.Exception
+        def e = thrown java.lang.Exception
         and: 'it contains some info about the problem'
-            e.message.contains('Please specify at least one inputFile in your docToolchainConfig.groovy')
+        e.message.contains('Please specify at least one inputFile in your docToolchainConfig.groovy')
     }
-
 }
--- a/changelog.adoc
+++ b/changelog.adoc
@@ -20,6 +20,7 @@ and this project tries to adhere to https://semver.org/spec/v2.0.0.html[Semantic
 * various fixes in `dtcw`, `dtcw.ps1`:
 ** pick the right environment if none provided by the user
 ** support of JAVA_HOME which was silently ignored.
+* https://github.com/docToolchain/docToolchain/issues/220[#220 convertToDocx and convertToEpub not working]
 
 === added
 
--- a/scripts/AsciiDocBasics.gradle
+++ b/scripts/AsciiDocBasics.gradle
@@ -220,9 +220,7 @@ task generateHTML (
         backends = ['html5']
     }
 
-    def sourceFilesHTML = sourceFiles.findAll {
-                                'html' in it.formats
-                            }
+    def sourceFilesHTML = findSourceFilesByType(['html'])
 //    onlyIf {
 //        sourceFilesHTML
 //    }
@@ -253,16 +251,14 @@ task generateHTML (
         }
     }
 
-doFirst {
-    if (sourceFilesHTML.size()==0) {
-        throw new Exception ("""
-
->> No source files defined for type HTML.
->> Please specify at least one inputFile in your docToolchainConfig.groovy
-""")
+    doFirst {
+        if (sourceFilesHTML.size()==0) {
+            throw new Exception ("""
+            >> No source files defined for type 'html'.
+            >> Please specify at least one inputFile in your docToolchainConfig.groovy
+            """)
+        }
     }
-
-}
 }
 //end::generateHTML[]
 
@@ -284,9 +280,7 @@ task generatePDF (
         'imagesoutdir'        : file("${docDir}/${config.outputPath}/images/").path
     )
 
-    def sourceFilesPDF = sourceFiles.findAll {
-        'pdf' in it.formats
-    }
+    def sourceFilesPDF = findSourceFilesByType(['pdf'])
 //    onlyIf {
 //        sourceFilesPDF
 //    }
@@ -309,6 +303,15 @@ task generatePDF (
         backends = ['pdf']
     }
 
+    doFirst {
+        if (sourceFilesPDF.size()==0) {
+            throw new Exception ("""
+            >> No source files defined for type 'pdf'.
+            >> Please specify at least one inputFile in your docToolchainConfig.groovy
+            """)
+        }
+    }
+
     /**
     //check if a remote pdfTheme is defined
     def pdfTheme = System.getenv('DTC_PDFTHEME')
@@ -318,16 +321,6 @@ task generatePDF (
         //TODO: finish this...
     }
     **/
-    doFirst {
-            if (sourceFilesPDF.size()==0) {
-                throw new Exception ("""
-
-    >> No source files defined for type PDF.
-    >> Please specify at least one inputFile in your docToolchainConfig.groovy
-    """)
-            }
-
-    }
 }
 //end::generatePDF[]
 
@@ -337,9 +330,7 @@ task generateDocbook (
         group: 'docToolchain',
         description: 'use docbook as asciidoc backend') {
 
-    def sourceFilesDOCBOOK = sourceFiles.findAll {
-        'docbook' in it.formats
-    }
+    def sourceFilesDOCBOOK = findSourceFilesByType(['docbook', 'epub', 'docx'])
 //    onlyIf {
 //        sourceFilesDOCBOOK
 //    }
@@ -365,21 +356,19 @@ task generateDocbook (
     doFirst {
         if (sourceFilesDOCBOOK.size()==0) {
             throw new Exception ("""
-
-    >> No source files defined for type docbook.
-    >> Please specify at least one inputFile in your docToolchainConfig.groovy
-    """)
+            >> No source files defined for type of '[docbook, epub, docx]'.
+            >> Please specify at least one inputFile in your docToolchainConfig.groovy
+            """)
         }
-
     }
 }
 //end::generateDocbook[]
 
 //tag::generateDeck[]
 task generateDeck (
-        type: AsciidoctorJRevealJSTask,
-        group: 'docToolchain',
-        description: 'use revealJs as asciidoc backend to create a presentation') {
+    type: AsciidoctorJRevealJSTask,
+    group: 'docToolchain',
+    description: 'use revealJs as asciidoc backend to create a presentation') {
 
     // corresponding Asciidoctor reveal.js config
     // :revealjs_theme:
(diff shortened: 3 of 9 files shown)


src/test/groovy/docToolchain/GenerateDeckSpec.groovy
```diff
@@ -31,10 +31,11 @@ class GenerateDeckSpec extends Specification {
...
                     .build()
-        then: 'we get an exception'
-            def e = thrown java.lang.Exception
-        and: 'it contains some info about the problem'
-            e.message.contains('Please specify at least one inputFile in your docToolchainConfig.groovy')
+        then: 'we get a warning'
+            println result.output
+            result.output.contains('WARNING: No source files defined for type revealjs. Converting all files')
+        then: 'the task still completes'
+            result.task(":generateDeck").outcome == SUCCESS
```
mh182: Is this really what we want?

If I called `generateDeck` but I have no file configured (or made a mistake configuring my files), I would expect to get an error, and an exit code different from 0.

As far as I understand, #220 want to convert the source files to `docx` or `epub` which implies a generated `docbook`. Converting the `docbook`format to the desired output format needs pandoc.

The docker container contains now pandoc (#433), but the local and sdk environment doesn't always have pandoc installed.

Explicit is better than implicit: this would mean we would not try to implicitly call pandoc (if not configured), but provide an error message which guides the user to add `docbook` to the converted formats.

In case the user uses a local environment where pandoc is not installed, we show an error message that the user has install pandoc (to conert do epub/docx).

PacoVK: You points are valid, but as i understood the discussion in #220 and what i would expect from a user perspective i think we should be less strict here. 

1. As per [comment](https://github.com/docToolchain/docToolchain/issues/220#issuecomment-433603052) - Not all files in a project are suitable for all output formats. However, i think it could be tedious to enforce explicit config here, although most tasks are suitable for any project file. Warning would be better suited here. 
2. If there is some dependency on the system missing, like pandoc, the correct config would not solve the issue, so the error will still remain, regardless of your configuration. 
3. One-Off tasks (e.g. you only want to run `generatePDF` Task once for a special use case) enforces the user with explicit config to configure the output format to `pdf` although the normal use case of the project does not need such.

Open for other opinions here :) @rdmueller  as you already were involved in #220 maybe you have something to add here?

mh182: > You points are valid, but as i understood the discussion in #220 and what i would expect from a user perspective i think we should be less strict here.
> 
>     1. As per [comment](https://github.com/docToolchain/docToolchain/issues/220#issuecomment-433603052) - Not all files in a project are suitable for all output formats. However, i think it could be tedious to enforce explicit config here, although most tasks are suitable for any project file. Warning would be better suited here.

As far as I understood, the statement "Not all files in a project are suitable for all output formats" meant "you forgot to create docbook before creating the docx/epub output".  I would expect the user provides a configuration for the different output types. So for example, if you want a slide deck and a manual, you would provide the following:

```
inputFiles = [
        [file: 'manual.adoc',       formats: ['html','pdf','docbook','docx','epub']],
        [file: 'my-presentation.adoc',       formats: ['revealjs']],
]
```

To generate the manual in EPUB format the user would have to make sure to call `generateDocbook` before calling `convertToEpub` (not sure how the output of the previous step can be used as input for the second one).

My main problem with warnings (more concrete, an exit code of SUCCESS) is that they are ignored. If we introduce a bug in `generateDeck` which would screw up the "file input selection" the user would get a warning but no output. Let's assume a user uses a CI build to create a web-page with her/his silde decks. Suddenly, all the slide decks on the home page would disappear without the user noticing.
 
>     2. If there is some dependency on the system missing, like pandoc, the correct config would not solve the issue, so the error will still remain, regardless of your configuration.

That's true. I was just explaining a use case which is closely related to the "configuration is missing" use case. Ignore this for the moment.

>     3. One-Off tasks (e.g. you only want to run `generatePDF` Task once for a special use case) enforces the user with explicit config to configure the output format to `pdf` although the normal use case of the project does not need such.

One-off tasks can be configured with `-P` option. I'm not sure how well this can be done (or if all tasks support parameter overriding). A concrete example would be nice to go into detail.

I see the problem of a silent failing CI build a bigger problem than the support of on-off task. But maybe we can find a solution for both.

PacoVK: Okay, i see your points. 
To come up with a solution here, i will rollback to the current failing mode. I would suggest to make tasks that are inevitably dependent on other tasks explicit using Gradles `dependsOn` 
Do you know of Tasks other than 
* 'docx','epub' depend on 'docbook'?

rdmueller: I think `publishToCOnfluence` depends on `generateHTML`

PacoVK: i changed the implementation accordingly. `dependsOn` was already there but docToolchain can now try to execute tasks on best effort

mh182: Maybe I’m a bit pedantic, but wouldn't this be a perfect candidate for a test case?

To sign off the changes, I could try out the different scenarios manually.

Reading through the patch, my guess for the current logic is:
- if an `inputFile` with `epub` (or `docx`) is defined, but no `docbook`, add the `docbook` implicitly and create everything (which may fail if pandoc is not installed)
- throw error
  - if someone calls `generateHTML` but no `inputFiles` with `html` format exists
  - if someone calls `generatePDF` but no `inputFiles` with `pdf` format exists
  - if someone calls `generateDocbook` but none of `inputFiles` has a `docbook, epub, docx` format
  - if someone calls `generateDeck` but no `inputFiles` with `revealjs` format exists (for that, we have a test)

Is my Gradle code reading correct?

I still think we should avoid doing too much "magic" behind the scenes. For the first use case (epub/docx but no docbook), would it not make more sense to abort with an error message and tell the user he/she has to add `docbook` to the output formats since `epub` requires a `docbook`?

This would also solve the problem in the case the user has no pandoc installed and suddenly is confronted with the error when calling `epub` docToolchain aborts with a cryptic error of  "A problem occurred starting process 'command 'pandoc''. How should she/he know that `adoc` --> `epub` in reality is `adoc` --> `docbook` -(pandoc)-> `epub`? 

The current error workflow for a user who wants to create an `epub` could just be improved with more descriptive error messages:
- `epub` but no `docbook` specified (note: I agree that the error description should be improved)
```
➜ ./dtcw convertToEpub
> Task :generateDocbook FAILED

* Where:
Script '/home/max/.doctoolchain/docToolchain-latest/scripts/AsciiDocBasics.gradle' line: 367

* What went wrong:
Execution failed for task ':generateDocbook'.
> java.lang.Exception: 
  
      >> No source files defined for type docbook.
      >> Please specify at least one inputFile 
```
- add the `docbook` format to the configuration and call it again (once again, the error description is not really helping)
```
➜ ./dtcw convertToEpub
> Task :generateDocbook
Converting /home/max/work/dtc/test-dtcw-2.3.0/src/docs/arc42/arc42.adoc

> Task :convertToEpub FAILED

* What went wrong:
Execution failed for task ':convertToEpub'.
> A problem occurred starting process 'command 'pandoc''
```

As a newbie user, I would be surprised to have the call to `:generateDocbook` since it is not configured anywhere.

In summary, my feedback for the current PR:
- we should write the test cases
- we should improve the error output
- I would prefer the explicit definition of `docbook` (which also makes code less complicate). But if you both think the implicit call to `docbook` is OK, I will not oppose it.

Sorry for the long reply (in the meantime I could have written the tests *lol*)

PacoVK: :D i really appreciate your long comment and i am totaly aligned to that. I think we shold adhere to get the test coverage even better. However, i just wanted to verify that we are all aligned on the very solution. But i guess we are getting there.
* I will provide a task that does some pre-check for pandoc to fail friendly 
* I will improve the error handling (initially thought about that, but that could have been a separate issue. I am okay doing this in the scope of that issue) 
* I do disagree on explicit `docbook` config, since imho this is implementation detail that should not be leaked to the user, but i understood you are fine with that. Hence, i will keep this mechanism as propsed. 
* ofc i will add some tests, and also extend some existing to test the failure mode

PacoVK: @mh182 @rdmueller  could one of you please checkout 'verifyPandoc' task on windows? 
This PR is getting quite huge, i also fixed a bug on the pandoc tasks where it was not possible to convert files when they were nested into directories. 

This PR will also close #350 

Still missing: I will improve the error output

PacoVK: After thinking a bit over the error output improvement, i think it's basically good enough for now to keep the scope here small. 
We should rather think about an extra branch that may refactor Error handling and may introduce docToolchain Exceptions, that leaves us more control and solution space of unified and meaningful error messages

mh182: > @mh182 @rdmueller could one of you please checkout 'verifyPandoc' task on windows? This PR is getting quite huge, i also fixed a bug on the pandoc tasks where it was not possible to convert files when they were nested into directories.

Sorry, I don't have an MS Windows environment to test.

> After thinking a bit over the error output improvement, i think it's basically good enough for now to keep the scope here small.
> We should rather think about an extra branch that may refactor Error handling and may introduce docToolchain Exceptions, that leaves us more control and solution space of unified and meaningful error messages

I agree with you. We should avoid too many changes before the release. We should focus on making the transition to the upgraded Java and plugins as smooth as possible.

mh182 (approved):
Looks good to me.

Rebase, squash and merge.

PacoVK:
@mh182  did we somehow verify `verifyPandoc` works on windows? I have no windows here

PacoVK:
I just could verify it 🍾

Issue #723, opened 2021-12-14
Dtcw Version 


Hellmy:
The dtcw version on the ng branch is 0.22 but the downloaded version is 0.27

Why is the current version not in the ng branch?

rdmueller:
I do understand that this is confusing, but I haven't found a better solution yet:

the home of dtcw is https://github.com/docToolchain/docToolchain.github.io . This way, the download link is `doctoolchain.github.io/dtcw` or even shorter `doctoolchain.org/dtcw`.

However, dtcw is also used to build the docToolchain website and therefore a copy of it is located in the main repository.

Any idea how we can make this less confusing?

Hellmy:
Is it really necessary to have the dtcw in the other repository? Only for the short url or anything else? Because from my understandig you could also copy the dtcw to the gh-pages of the docTollchain repository but the url would be longer...

silverdonkey:
"However, dtcw is also used to build the docToolchain website and therefore a copy of it is located in the main repository."

=> In the build you could first download the 'dtcw' from 'doctoolchain.org/dtcw' and then run it ? This way no copy would be needed...

Hellmy:
Also a interesting option - possible only a dtcw.adoc file with the link and a short description

mh182:
As Ralf mentioned, the versioning discrepancy of `dtcw` and used docToolchain version (see #1030, respective #1049) stem from the release process which is not as mature as we would like to have it.

The GHA workflow [update-dtcw.yaml](https://github.com/docToolchain/doctoolchain.github.io/blob/master/.github/workflows/update-dtcw.yml) from the https://github.com/docToolchain/doctoolchain.github.io repo copies the shell scripts from https://github.com/docToolchain/docToolchain into the repo to releases the scripts.

So the source of `dtcw` (the one we develop on) is located in https://github.com/docToolchain/docToolchain. But the user downloads the released `dtcw` script from the docToolchain homepage. 

From a security point of view I would keep it this way since a normal docToolchain user comes from https://doctoolchain.org/ and thus associates the domain with the tool (I created #1106 since currently the installation instructions use `doctoolchain.github.io`).

I would suggest the following improvement:
- we add string templates in `dtcw` in https://github.com/docToolchain/docToolchain for
    - the default docToolchain version (`$DTC_VERSION`)
    - the git hash from which `dtcw` was released (copied over)
[...]

PR #1140, opened 2023-05-01
dtcw: show git hash for docToolchain latest version

### LINKED ISSUE #1139: DTC_VERSION=latest dtwc should show docToolchain git hash
To facilitate the creation of bug reports when working with the `latest` version (or `latestdev`) we should output the git hash of docToolchain on invoking `dtcw`.

```
DTC_VERSION=latest dtcw --version
dtcw 0.50 - ##DTCW_GIT_HASH##
docToolchain 2.2.1 - 07b0a8a7
OS/arch: Linux/x86_64
```

For a typical example of such a bug report is #1136 (tests before release).

mh182:


rdmueller:
so, we only output the git hash for development version? I thought we alsways output it / generate it when we deploy it?

rdmueller (approved):
looks fine for me

PacoVK (approved):

mh182:
> so, we only output the git hash for development version? I thought we alsways output it / generate it when we deploy it?

This is the git hash for the docToolchain, not `dtcw`, and not the released docToolchain. So if someone uses `DTC_VERSION=latest` we get output information of the git repo on which the bug report was created.

#723 will retrieve the git hash for which `dtcw` is released.

Later we still can decide if we add the git hash for docToolchain on a release version (which should not be neccessary since we know for which git hash a release is).

PR #1143, opened 2023-05-01
fix (closes #829): make docker container name unique to avoid conflicts

### LINKED ISSUE #829: Multiple run of dtcw with docker is failing because of a container name conflict
**Describe the bug**
Multiple run of docToolchain via docker is not supported on the same machine

**To Reproduce**
Have long run of docToolchain and start 2 of them in parallel.

**Current behavior**

Failing with:
```
./dtcw generateSite
dtcw - docToolchain wrapper V0.29
docToolchain V2.0.3
docker available
/usr/bin/docker
use docker installation
docker: Error response from daemon: Conflict. The container name "/doctoolchain" is already in use by container "3352c81b6369435e454b2343e3862cb5d14c151fad635322ec9b505909065270". You have to remove (or rename) that container to be able to reuse that name.
See 'docker run --help'.
```

**Expected behavior**
Is it really required to name the container to a fix name, couldn't be something like "doctoolchain+<currentTimeMillis>" or something random.

**Configuration**
 - docToolchain 2.0.3
 - OS: Linux

mh182:


PacoVK (changes requested):
Looks good, but what about the [powershell script](https://github.com/mh182/docToolchain/blob/829_container_name_conflict/dtcw.ps1#L547)?

mh182:
> Looks good, but what about the [powershell script](https://github.com/mh182/docToolchain/blob/829_container_name_conflict/dtcw.ps1#L547)?

You are right, I forgot about that part.

I have to take a look if I can test the PowerShell script on my Linux machine. If that works, I'm going to adapt it.

PacoVK (approved):
LGTM 👍

PR #1146, opened 2023-05-02
doc: update tutorial to create presentations with reveal.js

### LINKED ISSUE #869: generateDeck: tutorial for revealjs should be updated
thanx to @wonderbird , we now have a tutorial on how to use the generateDeck task.

see #861 for more details.

but there is more work to do on this:

the tutorial is based on the docs-as-co.de tutorial which is outdated - still works but outdated.

- [ ] ToDo: remove docs-as-co.de tutorial

the ./dtcw version of docToolchain does not come with reveal.js . The installer does not clone the corresponding repository. We should fix this

- [x] fix clone of  reveal.js repository for installer

since a few days, we don't need sdk man anymore to install java for docToolchain. a `./dtcw getJava` is enough

- [ ] update tutorial accordingly

mh182:
Remove download of arc42 template, which already has its own tutorial.

This PR closes #869.


mh182:
Is there a way to auto-update the slide deck when I'm working on the deck?

If yes, I would add this to the tutorial to show a working round trip.

rdmueller:
no, no auto-update yet.

rdmueller (approved):
Thanx, I like the new tutorial.

I only wonder about the use of `xdg-open`. This command will not work on all systems. Should we just say "open file xyz in your browser"?

src/docs/020_tutorial/110_presentations_with_revealjs.adoc
```diff
@@ -0,0 +1,128 @@
...
+The generated configuration file contains the configuration for all tasks provided by _docToolchain_, which may be overwhelming.
+====
+
+- Generate the slide deck
+
+```shell
+$ ./dtcw generateDeck
+```
+
+- Open your slide deck in your browser
```
PacoVK:
```suggestion
- Find your slide deck at `build/decks/presentation.html` and open it in your browser
```

PacoVK: also agree with @rdmueller , we should remove `xdg-open` as it may not be available on the system :) Very nice tutorial!

mh182: I'm going to change it as requested. My point of view is too much Linux centric.

The main pain point is that the user is not notified where the generated output is going to be located. I created  #1151 to address this.

PR #1148, opened 2023-05-03
dtcw: install java 17

### LINKED ISSUE #1129: `dtcw install java` installs Java 17
Let's make Java 17 the default major version.

`./dtcw install java` installs Java 17.

Check and update the documentation in case we use older Java versions.

mh182:
Closes #1129

### DIFF
--- a/.github/workflows/dtcw-tests.yaml
+++ b/.github/workflows/dtcw-tests.yaml
@@ -33,7 +33,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest]
       fail-fast: false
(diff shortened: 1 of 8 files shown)


mh182:
Hmm, the deployment on Netlify breaks since we use docToolchain 2.2.1. 

Should we not use the `latest` version when we deploy the page?

rdmueller:
yes, I guess latest makes sense... I will switch this in a moment

rdmueller:
the script tried to switch the version to "dev". I have just fixed the expressions. let's see if the preview now deploys

rdmueller (approved):
looks good to me!

PacoVK (approved):

PR #1162, opened 2023-05-09
Record page anchor before header removal. fixes #1161

### LINKED ISSUE #1161: publishToConfluence looses the id when generating level 2 page anchors
**Describe the bug**
When using the `subpagesForSections=2` setting, Confluence will loose the links to level 3 adoc headers

**To Reproduce**
1. Set the `concluence.with` setting for `subpagesForSections` to 2
2. Include a level 3 heading on an included page (In my case a level 2 heading, but with leveloffset +1, using the https://docs.asciidoctor.org/pdf-converter/latest/interdocument-xrefs/#converting-interdocument-xrefs-to-internal-xrefs syntax)
3. The generated HTML page will render those as headers as h3 headings.
4. On publishing, the pageAnchor for these headings gets lost, and no link macro will be generated for these links, resulting in broken links on the published Confluence page.

**Expected behavior**
The pageAnchors are not lost, and the links on the published Confluence page work :-)

**Configuration**
- docToolchain 2.2.1, as well as docToolchain latest.
- OS: Windows
- JDK 11

**Additional context**
I created a minimal example containing this issue: https://github.com/srotman/docToolchain-1161

According to my analysis, the problem lies in the `getPagesRecursive` function: https://github.com/docToolchain/docToolchain/blob/c9cc6f3a63c1ae57b586153e36a143ef843f2192/scripts/asciidoc2confluence.groovy#L944-L958
It seems that the `new Elements(...)` constructor (line 949) does not perform a deep Copy, resulting in a removal of the h3 element on line 950. The subsequent select of that same element on line 958 now yields no results any longer.

If this line: https://github.com/docToolchain/docToolchain/blob/c9cc6f3a63c1ae57b586153e36a143ef843f2192/scripts/asciidoc2confluence.groovy#L958 is moved up to line 945 (right after the initial select) the pageAnchors are kept correctly, and the Confluence links will work as expected.

srotman:
### All Submissions:

* [x] <s>Does your PR affect the documentation?</s>
* [x] <s>If yes, did you update the documentation or create an issue for updating it?</s>

### Your first submission

### DIFF
--- a/changelog.adoc
+++ b/changelog.adoc
@@ -17,6 +17,7 @@ and this project tries to adhere to https://semver.org/spec/v2.0.0.html[Semantic
 * https://github.com/docToolchain/docToolchain/issues/973[#973: `dtcw getJava` doesn't work without `local`]
 * https://github.com/docToolchain/docToolchain/issues/1109[#1109: docToolchain release notes contain releases twice]
 * https://github.com/docToolchain/docToolchain/issues/1031[#1031: dtcw ignores installed Java RE when docker is installed - your java version 17 is too new]
+* https://github.com/docToolchain/docToolchain/issues/1161[#1161: publishToConfluence looses the id when generating level 2 page anchors]
 * various fixes in `dtcw`, `dtcw.ps1`:
 ** pick the right environment if none provided by the user
 ** support of JAVA_HOME which was silently ignored.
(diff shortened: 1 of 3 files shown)


mh182:
@PacoVK Could you review this PR? 
You are more into the Gradle/Groovy code than I am. 
In an ideal world, we would need a test case to show the bug and the fix.

PacoVK:
thanks to @srotman  for the contribution :) Although i agree for the test case i think there is much work to do before we can write a test. Hence, i would like to postpone this and adress this later

PacoVK (approved):

PR #1174, opened 2023-05-12
introduce generateContent task

PacoVK:
# What this PR is about

It introduces a new task called `generateContent`. It's basically an integration to openAI. You can put comment blocks into your input file like 
```asciidoc
////
AI: write something short about habits
////
````
which will then be fully replaced by the text provided through openAI. 

_Disclaimer_: You'll need an API key from openAI in order to make this task work

### All Submissions:

### DIFF
--- a/Config.groovy
+++ b/Config.groovy
@@ -387,3 +387,15 @@ structurizr.with {
     }
 }
 //end::structurizrConfig[]
+
+//tag::openAIConfig[]
+// Configuration for openAI related tasks
+openAI = [:]
+
+openAI.with {
+    //token = ''
+    //model = "text-davinci-003"
+    //maxToken = '500'
+    //temperature = '0.3'
+}
+//end::openAIConfig[]
--- a/Config.groovy
+++ b/Config.groovy
@@ -393,7 +393,10 @@ structurizr.with {
 openAI = [:]
 
 openAI.with {
-    //token = ''
+    // This task requires a person access token for openAI.
+    // Ensure to pass this token as parameters when calling the task
+    // using -PopenAI.token=xx-xxxxxxxxxxxxxx
+
     //model = "text-davinci-003"
     //maxToken = '500'
     //temperature = '0.3'
--- a/scripts/generateContent.gradle
+++ b/scripts/generateContent.gradle
@@ -22,11 +22,11 @@ task generateContent (
     group: 'docToolchain'
 ) {
     doLast{
-        def token  =  (findProperty("openAI.token")?:config.openAI.token)?:null
+        def token  =  findProperty("openAI.token")?:null
         if(token == null){
             throw new Exception ("""
             >> No Token provided for openAI.
-            >> Please specify the token using -PopenAI.token=xx-xxxxxxxxxxxxxxxx or in your docToolchainConfig.groovy
+            >> Please specify the token passing the parameter -PopenAI.token=xx-xxxxxxxxxxxxxxxx
             """)
         }
         def model = (findProperty("openAI.model")?:config.openAI.model)?:"text-davinci-003"
--- a/scripts/generateContent.gradle
+++ b/scripts/generateContent.gradle
@@ -1,0 +1,74 @@
+buildscript {
+    repositories {
+        maven {
+            credentials {
+                username mavenUsername
+                password mavenPassword
+            }
+            url mavenRepository
+        }
+    }
+    dependencies {
+        classpath 'com.theokanning.openai-gpt3-java:service:0.12.0'
+    }
+}
+
+import com.theokanning.openai.service.OpenAiService;
+import com.theokanning.openai.completion.CompletionRequest;
+
+//tag::generateContent[]
+task generateContent (
+    description: 'Generate documentation with the help of OpenAI',
+    group: 'docToolchain'
+) {
+    doLast{
+        def token  =  (findProperty("openAI.token")?:config.openAI.token)?:null
+        if(token == null){
+            throw new Exception ("""
+            >> No Token provided for openAI.
+            >> Please specify the token using -PopenAI.token=xx-xxxxxxxxxxxxxxxx or in your docToolchainConfig.groovy
+            """)
+        }
+        def model = (findProperty("openAI.model")?:config.openAI.model)?:"text-davinci-003"
+        def maxToken = (findProperty("openAI.maxToken")?:config.openAI.maxToken)?:500
+        def temperature = (findProperty("openAI.temperature")?:config.openAI.temperature)?:0.3
+        logger.info "docToolchain> Generating AI content"
+        logger.info "docToolchain> using model ${model}"
+        logger.info "docToolchain> limiting output to ${maxToken} tokens"
+
+        def pattern = /\/\/\/\/[\r\n\s]AI:*(.*?)[\r\n\s]*\/\/\/\// // match asciidoc block comments
+        def service = new OpenAiService(token, Duration.ofSeconds(60));
+        def sourceFiles = config.inputFiles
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: Config.groovy [hunk 1]
```diff
@@ -383,3 +383,14 @@ structurizr.with {
...
 //end::structurizrConfig[]
+
+//tag::openAIConfig[]
+// Configuration for openAI related tasks
+openAI = [:]
+
+openAI.with {
+    //token = ''
+    //model = "text-davinci-003"
+    //maxToken = '500'
```
rdmueller, line 401: shouldn't we also add the temperature as parameter`?

rdmueller (commented):
Great! I like the idea to use openAI as writing buddy. I also think this is an interesting start.
I wonder if we could make it more interactive and use more priming.

For instance, it could make sense to add some primer to the config like this (which I often use):

```
when you create AsciiDoc output
- format it as code (so I can copy and paste it)
- use '=' for headlines
- start with level '==' as headline
- write each sentence or half-sentence in a new line
- use attributes for URLs and define them at the start of the document
```

It could also make sense to use the `AI`-statements in a dialog/session, so that one generated text block build upon the last ones... (or maybe is this already implemented? Didn't check the source)

Config.groovy
PacoVK: sure, it makes sense. However, we should think about sensible defaults for any parameter to ease life for users that are not familiar with that terms 👍

PacoVK:
Yes, indeed i also thought about having a more interactive implementation, but was not really sure about the capabilities with Gradle here. 
The current task is quite bare-bones but if we know how the "writing buddy" could look like, we can improve the user experience here. 
It would also be awesome to let the user decide whether the openAI suggestion is acceptable or if the user wants to tweak the given input.

Config.groovy
rdmueller: If we use 0.0 as default, results will be reproducible

PacoVK: wouldn't that prevent the possibillity to regenerate the result if it is not accurate enough? 0.0 means the user would have to change the given input

mh182:
What I really would need is the feature to rephrase existing paragraphs. Would this also work with this task?

If yes, could you add an example how this could be done?

Checking the task documentation, I asked myself, "Why do we show the task's source code?". The target group of the task documentation is a user who usually doesn't need to know about the internals.

IMHO it makes the page unnecessarily long and harder to digest.

rdmueller:
@mh182 

> IMHO it makes the page unnecessarily long and harder to digest.

yes, but another topic. My goal is to remove the source code and hide it behind a link or tab

rdmueller:
> What I really would need is the feature to...

I think we should consider this as experimental and I guess we soon will have plugins for our IDEs etc. to solve these feature problems. Let's give it a try and see how this feature evolves...

scripts/generateContent.gradle
```diff
@@ -0,0 +1,74 @@
...
+import com.theokanning.openai.service.OpenAiService;
+import com.theokanning.openai.completion.CompletionRequest;
+
+//tag::generateContent[]
+task generateContent (
+    description: 'Generate documentation with the help of OpenAI',
+    group: 'docToolchain'
+) {
+    doLast{
+        def token  =  config.openAI.token?:null
```
mh182: I think it would make sense to provide the means so that configuration parameters can be provided via command line. 
This way the user doesn't have to change the configuration files in case they want to call the task just once.

```
def token = (findProperty("openAI.token")?:config.openAI.token)?:null
```

Config.groovy
rdmueller: 0.0 means: give me the statistically best answer. I think this would be ok. But I don't mind if we keep 0.3 (or anything else), because I think we have to give it a try.

rdmueller (approved):
I want to see this in action...

mh182 (changes requested):

scripts/generateContent.gradle
PacoVK: good chatch, thanks!

scripts/generateContent.gradle
```diff
@@ -0,0 +1,74 @@
...
+        def token  =  (findProperty("openAI.token")?:config.openAI.token)?:null
+        if(token == null){
+            throw new Exception ("""
+            >> No Token provided for openAI.
+            >> Please specify the token using -PopenAI.token=xx-xxxxxxxxxxxxxxxx or in your docToolchainConfig.groovy
+            """)
+        }
+        def model = config.openAI.model?:"text-davinci-003"
+        def maxToken = config.openAI.maxToken?:500
+        def temperature = config.openAI.temperature?:0.3
```
mh182: Would it not make sense to provide means to overrule all configure parameters via command line?
Trying out stuff is less painful.

I know that not all tasks provide this feature. But we want to go into this direction. So let's make new tasks as we would like to have them

mh182 (changes requested):
Good work. I like it.

Apart from passing all configuration parameters through to the command line.

scripts/generateContent.gradle
PacoVK: I initially thought about seperating config values that are required to even execute the task, like the token, and config params that ay be more advanced settings like these. 
I'm fine if we default to parameter overwrite always possible. This way we could have a consistent configuration mechanism that follows precedence like:
1. via CLI (always win)
2.  via config file
3. default value

PacoVK:
Concerning the implementation details as part of the docs, i started a discussion, to find out how we can handle this in the future https://github.com/docToolchain/docToolchain/discussions/1175

PacoVK:
Hm, somehow the DtcwOnPowershellSpec is broken on Java 17... i have no glue tbh it just stopped working :D Does this Test run intentionally on Linux environment?

rdmueller:
Yes, it intentionally runs on linux with powershell installed

PacoVK:
> Yes, it intentionally runs on linux with powershell installed

Ah, then there is this part missing in the `default-build.yml` right? 
In the `dtcw.ps1-tests.yaml` there is 

```
      - name: install powershell
        shell: bash
        run: |
          if [ "$RUNNER_OS" == "Windows" ]; then
            echo "pwsh already installed"
          else  
            sudo apt-get install -y wget apt-transport-https software-properties-common
            wget -q "https://packages.microsoft.com/config/ubuntu/$(lsb_release -rs)/packages-microsoft-prod.deb"
            sudo dpkg -i packages-microsoft-prod.deb
            rm packages-microsoft-prod.deb
            sudo apt-get update
            sudo apt-get install -y powershell
          fi
```

PacoVK:
i am super confused about the behaviour of the test suite and how it is breaking

rdmueller:
I guess you are right. let me fix this...

rdmueller:
seems to work now

mh182:
The patch is fine. 

I have one "minor concern" about  the configuration option `openAI.token`. As far as I understood this would be the user's personal OpenAI token.

Do we really want to nudge the user to put this in the configuration file which will end up in the revision control system?

When collaborating in a team, most probably each user has its own API key.

My suggestion would be to make the `openAI.token` a parameter which is provided by the command line only. Not within the configuration template.

mh182 (approved):

mh182:
@PacoVK Would it make sense to write a test for this task?

The problem I see here is which API key do we use to run the test. Or the test would mock out the call to the OpenAI API and test the rest.

PacoVK:
> Would it make sense to write a test for this task?

Yes, it always makes sense to test :D But as you figured out this would mean to have a test token and i did not yet discover a open source program on openAI site. 
Mocking would mean implementation refactoring into groovy files. I would not open this box as well here, but should definitely be a follow-up.

Issue #1107, opened 2023-04-12
Provide a template for a docToolchain task test with Spock


mh182:
docToolchain uses the [Spock Testframwork](https://spockframework.org/) to test tasks. The tests are located at `src/test/groovy/docToolchain`.

If a new developer on-boards on the project, we would have a hard time to provide a guideline how to write tests. 

The main problem I had was that I could not "detect" (or identify) the test in the test folder
```
src/test/
├── build
│   ├── exportConfluenceSpec
│   │   ├── AsciiDoc_Test_Home
│   │   └── images
│   └── test
│       └── docs
├── docs
│   ├── excel
│   │   └── Sample.xlsx
│   └── images
│       └── ea
├── groovy
│   ├── docToolchain
│   │   └── export2confluence
│   └── site
├── testAsciidoctor
│   └── docs
├── testConfluenceSpace
│   └── attachments
│       ├── 2039316482
│       └── 2039447553
├── testData
│   ├── excel
│   │   └── Sample.xlsx
│   └── structurizr
│       ├── default
│       └── filename
├── testPdfDefaultTheme
│   ├── pdfTheme
│   └── testPdfDefaultThemeDocs
│       ├── ea
│       └── images
└── testPdfSpecificTheme
    ├── testPdfSpecificThemeDocs
    │   ├── ea
    │   └── images
    └── testPdfTheme
```

In an ideal world, we would have the test code near the unit-under-test. But before we can move the test code closer to the task, we should decide on how we organize/structure our tests.

The goal of this ticket is to decide how we structure a test case on the file system.
- where do we put the input files
- where do we put the output files (so we can clean up a test run easily)
- where do we put the configuration file
- test fixtures
    - how do we structure tests for a task if more than one test setup is needed
    - how do we structure tests if a test fixture is shared between tests of different tasks
- common patterns and naming convention for a test

The test(s) of a task should be self-contained. Meaning that we could remove the tests of a task by just deleting a folder.

PR #1178, opened 2023-05-17
exportOpenApi: provide configuration parameters via command line

### LINKED ISSUE #1179: Improve exportOpenApi task for easier integration into master documents
Make the task `exportOpenApi` more flexible so the OpenAPI specification in Ascidoc format can be integrated into existing documentation in Asciidoc format (for example into the arc42 template).

The proposed changes are based the [comment on how to integrate OpenApi specification into ar42 documentation](https://github.com/docToolchain/docToolchain/discussions/1177#discussioncomment-5947082) 

- The OpenAPI specification file can be provided via command line with `-P openApi.specFile=some-path/to/the-spec.yaml`
- The location/name of the generated Asciidoc format of the OpenAPI specification can be configured. This should be optional. If not provided, create the file in `${targetDir}/OpenAPI`.
- Instead of providing `infoEmail` and `infoUrl` the user can configure `additionalProperties`. This way the export provides more flexibility.
- Provide output on console where the Asciidoc file was created.

The proposed change to the configuration for the `exportOpenApi` would look something like

```
openApi = [:]

openApi.with {
    specFile = 'src/docs/petstore-v2.0.yaml'

    // outputDir = "${targetDir}/OpenAPI"        // optional

    //additionalProperties = [                                 // optional
    //    infoUrl = 'https://my-api.example.org'
    //    infoEmail = 'info@example.org'
    //    headerAttributes = false
    //]
}
```

**Describe alternatives you've considered**
None. But I would be glad if someone would come up with something else.

**Additional context**
This issue will not address #1167 but we should discuss how we could configure the export of more than one OpenAPI specification.

mh182:
Changes in `exportOpenApi`: the property `openApi.specFile` may be provided via the command line with `-PopenApi.specFile`. This would ease the use case of having more than one OpenAPI specification file, as described in #1167.

The `exportOpenApi` task did not have a test case yet. So I try to gather feedback about how test case should be structured  (see #1107).

Test case structure on the file system:
- We use the [Spock testing and specification framework](https://spockframework.org/) to test the behavior of the docToolchain tasks.
- The unit tests of a docToolchain task are located in `src/test/groovy/scripts/<task_name>`.  
    - Reasoning: Groovy unit tests are by convention located in `src/test/groovy`. The content of this directory mirrors the structure and location of SUT, the system-under-test (i.e. the docToolchain tasks located in `scripts`)
- The file name providing the test code matches the test class name: `<task_name>Spec` (starting with Uppercase)
    - Reasoning: that's how it is now, so why change it?
- The tests are isolated: this means all necessary test code and data (files), except for the SUT itself, are located in its directory. Note: the **output generated by the tests** are generated within the test directory.
    - Reasoning: 
        - it facilitates the creation of a task test template fore new tasks (`cerateTask`)
        - it is easier for developers to understand what the minimal data for the test case is
        - it prevents sharing of test data, which increases the likelihood for brittle tests
        - in case a task is removed, its test case can easily be removed as well
- Test configuration file should only contain the minimal necessary information. By default, it should be called `config.groovy`. The test case can have more than one configuration file (`config-*.groovy`).
    - Reasoning: it's an easy pattern to separate configuration files from test data
- The structure on the file system of the test data should match the one in the real project

This PR includes the unit tests with the proposed structure for `exportOpenApi`.
```
src/test/groovy/scripts/
└── exportOpenApi
    ├── config.groovy
    ├── config-openApi-missing.groovy
    ├── ExportOpenApiSpec.groovy
    └── src
        ├── petstore-v2.0.yaml
        └── petstore-v3.0.yaml
```

- `ExportOpenApiSpec.groovy`: the Spock unit tests
- `config.groovy`: this is the configuration files used by most tests
- `config-openApi-missing.groovy`: a configuration file where the `openApi` configuration is missing
- `src/petstore-*.yaml`: the OpenApi specifications used by the test
- the Asciidoc file generated by the tests is created in `src/test/groovy/scripts/exportOpenApi/build`


scripts/exportOpenApi.gradle
```diff
@@ -14,19 +14,18 @@ buildscript {
...
         group: 'docToolchain',
         description: 'exports OpenAPI specification to the asciidoc file') {
 
+    def specFile = findProperty("openApi.specFile")?:config.openApi.specFile
     if (!specFile) {
-        logger.info("\n---> OpenAPI specification file not found in Config.groovy (https://doctoolchain.github.io/docToolchain/#_exportopenapi)")
-        return
-    } else {
-        logger.info("Found OpenAPI specification in Config.groovy")
+        throw new GradleException("Missing property 'openApi.specFile': please provide the location of the OpenAPI specification file")
```
mh182: These two lines break a lot of other tests. 
Can anyone explain me, why throwing an exception in `exportOpenApi` affects other tasks? As far as I can tell, none of those tasks depend on `exportOpenApi`.

PacoVK: Yes, take a look at the [Gradle build lifecycle](https://docs.gradle.org/current/userguide/build_lifecycle.html)

> A Gradle build has three distinct phases. Gradle runs these phases in order: first initialization, then configuration, and finally execution.

Hence other tests have no `specFile`, throwing an exception breaks the script in configuration phase. The logic in latest `ng` currently just emits an info log

A solution would be to wrap this logic into `doLast` or `doFirst` closure then it would only be executed in the Tasks execution phase

PacoVK:
Since this PR introduces a new file structure for tests to put them under `src/test/groovy/scripts` i would like to open a discussion here. 
1. Although putting tests under `src/test/groovy/scripts` folder reflects the current packaging of the scripts it still not following the standard conventions for Gradle projects. I would like to keep the current structure `src/test/groovy/doctoolchain` but start to move the actual `.groovy` files under `scripts` to the standard location, in that case `src/main/groovy/doctoolchain` --> This way IDEs won't struggle with importing docToolchain (currently manual setup is required to be able to link sources) 
2. A further step is to move some custom script logic in a `doLast` block (like `exportStructurizr`) into a seperate `.groovy` file. This way business logic will be separated from the actual Gradle specific logic and we can easily test in isolation. (an example would be `exportToConfluence`) The actual `.gradle` task files will look quite similar then.
3. Some tasks like `exportEA` even have inner classes in their `.gradle` build definition. Stuff like that makes testing in isolation impossible. 

TL;DR; My suggestion here is to adhere your basic concept which is good, but rather try to restructure the actual source files than try to adopt the test suite to the existing structure.

PacoVK:
One small remark on your current proposal: 

> src/petstore-*.yaml: the OpenApi specifications used by the test

I'd like to have them under 

> resources/petstore-*.yaml: the OpenApi specifications used by the test

because the `.yaml` files are test resources rather than source

mh182:
> Since this PR introduces a new file structure for tests to put them under `src/test/groovy/scripts` i would like to open a discussion here.
> 
>     1. Although putting tests under `src/test/groovy/scripts` folder reflects the current packaging of the scripts it still not following the standard conventions for Gradle projects. I would like to keep the current structure `src/test/groovy/doctoolchain` but start to move the actual `.groovy` files under `scripts` to the standard location, in that case `src/main/groovy/doctoolchain` --> This way IDEs won't struggle with importing docToolchain (currently manual setup is required to be able to link sources)

Good feedback. I do not know Gradle standard conventions. I will move the directory containing the test code into the existing `src/test/groovy/doctoolchain`. But I stand with the proposal, that each test has its own directory with the resources needed to run the test.

>     2. A further step is to move some custom script logic in a `doLast` block (like `exportStructurizr`) into a seperate `.groovy` file. This way business logic will be separated from the actual Gradle specific logic and we can easily test in isolation. (an example would be `exportToConfluence`) The actual `.gradle` task files will look quite similar then.

I agree with you. But this is out of scope for this ticket. The reason for this PR is to align us "how do we test tasks". The outcome would be a standard/guideline what minimal tests we want to have for new tasks.

>     3. Some tasks like `exportEA` even have inner classes in their `.gradle` build definition. Stuff like that makes testing in isolation impossible.

Once again, I agree with you. But I would not refactor the existing code without good test coverage.

> TL;DR; My suggestion here is to adhere your basic concept which is good, but rather try to restructure the actual source files than try to adopt the test suite to the existing structure.

We can restructure the source code when we have the test code in place ;-)

mh182:
> One small remark on your current proposal:
> 
> > src/petstore-*.yaml: the OpenApi specifications used by the test
> 
> I'd like to have them under
> 
> > resources/petstore-*.yaml: the OpenApi specifications used by the test
> 
> because the `.yaml` files are test resources rather than source

I disagree with you on this point. 

The main point of a test should be to reflect the structure of the documentation project structure. This means, following our standard project structure, if an OpenAPI specification is located in `src/docs/some-spec.yaml` I even would go so far as to put the spec in the test into `<test-directory>/src/docs`.

Having "stuff" in a directory "resources" is the "Gradle/Java" way of thinking, which doesn't necessarily reflect a documentation project directoyr structure.

PacoVK:
> The main point of a test should be to reflect the structure of the documentation project structure. This means, following our standard project structure, if an OpenAPI specification is located in src/docs/some-spec.yaml I even would go so far as to put the spec in the test into <test-directory>/src/docs.

Ah, yes makes sense, thanks for pointing that out. I agree if we align to adhere the documentation project directory structure.

PacoVK:
> But I would not refactor the existing code without good test coverage

This would just kill the scope of this PR i guess :D But if we align on the things discussed above, we could start incrementally.

src/test/groovy/scripts/exportOpenApi/src/petstore-v2.0.yaml
PacoVK, line 1: is this file still in use somewhere?

mh182: At the moment nowhere. 

I haven't finished this PR. I'm going to remove it.

scripts/exportOpenApi.gradle
```diff
@@ -14,38 +14,54 @@ buildscript {
...
+    // configuration phase. Defining those properties in doFirst() throws
+    // exception "propert 'X' is final and cannot be changed any further".
     generatorName = 'asciidoc'
-    outputDir = "${targetDir}/OpenAPI".toString()
     inputSpec = "${docDir}/${specFile}" // plugin is not able to find file if inputPath is defined as '.'
+    outputDir = "${targetDir}/OpenAPI".toString()
+
+    // Define cache is always out-of-date to enforce file generation regardless of input and output file
+    outputs.upToDateWhen { false }
+    outputs.cacheIf { false }
```
mh182, line 57: @PacoVK I would appreciate it, if you could take a look into the implementation of this task.

The implementation looks strange. Defining the `GenerateTask` properties in doFirst() throws exceptions. My Gradle mojo is below average. Maybe there is a better way.

scripts/exportOpenApi.gradle
```diff
@@ -14,38 +14,54 @@ buildscript {
...
+
     if (!specFile) {
-        logger.info("\n---> OpenAPI specification file not found in Config.groovy (https://doctoolchain.github.io/docToolchain/#_exportopenapi)")
+        // If the specFile is not defined skip the rest. Otherwise GenerateTask
+        // is called which fails with a misleading error message "file doesn't exist".
+        // We throw the descriptive "missing property" error message during execution phase.
         return
-    } else {
-        logger.info("Found OpenAPI specification in Config.groovy")
     }
```
mh182, line 46: This also looks strange. But I had to add this `return` statement otherwise `GenerateTask` would throw an exception and I never reached the code in doFirst() which contains a better error message.

PacoVK: need to investigate a bit, on the first glance it's because you use just a 3rd [partie task](https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator-gradle-plugin/src/main/kotlin/org/openapitools/generator/gradle/plugin/tasks/GenerateTask.kt) and the implementation is just offering this behaviour. Maybe there is something to turn off the validation

rdmueller (approved):
I am fine with the changes

PacoVK (approved):
Since this is already an improvement, i am also fine with that

rdmueller:
@PacoVK can you please take a look at the one conflict? I am not sure which is the right code:

image

raifdmueller:
Hi, and thank you for this pull request! 🙏

Let me be honest: in the past we simply didn't have the time for someone to give the open PRs the attention they deserve — and that's why this one has been sitting here untouched. We're sorry about that, because every PR represents real work and good intentions.

We're now trying to work through this backlog with the help of Claude Code. As part of that, we'll **go through each PR individually and decide whether it's still relevant or not**. Please bear with us if a PR ends up being closed — that's not a judgement on your work, it's just part of the cleanup.

**Important note:** We're currently working on **docToolchain v4**: https://doctoolchain.org/docToolchain/v4.0.x/ — so some changes may already be outdated or may need to be integrated differently.

If you'd like to support the project (time is our scarcest resource 😉), we'd really appreciate it via either of these links:
- https://opencollective.com/doctoolchain
- https://github.com/sponsors/rdmueller

Thanks again for contributing, and for your patience!

---
_Generated by [Claude Code](https://claude.ai/code)_

raifdmueller:
Closing this as part of the backlog cleanup. 🧹

The feature proposed here — providing `exportOpenApi` configuration (`openApi.specFile`) via the command line — is genuinely useful, but this PR has been inactive since Dec 2024 and now conflicts with the current `ng` branch. Since we're focused on **docToolchain v4** (https://doctoolchain.org/docToolchain/v4.0.x/), it makes more sense to re-implement it cleanly against the new architecture rather than untangle the conflicts here.

I've opened **#1648** to track the v4 re-implementation, explicitly referencing this PR (including your proposed Spock test structure for `exportOpenApi`, which is worth picking up again).

Thank you very much for the original work and especially for starting the discussion on how task test cases should be structured, @mh182 — that input wasn't lost, it's carried forward in #1648. 🙏

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1183, opened 2023-05-23
added req42 as template

rdmueller:
not perfect, but works

### DIFF
--- a/scripts/downloadTemplate.gradle
+++ b/scripts/downloadTemplate.gradle
@@ -12,7 +12,14 @@ task downloadTemplate(
             return new String((char) 27) + "[${colors[color]}m${text}" + new String((char) 27) + "[0m"
         }
         def lang, help
-        def templates = [arc42:"https://github.com/arc42/arc42-template/raw/master/dist/arc42-template-${->lang}-${->help}-asciidoc.zip"]
+        def templates = [
+            arc42:"https://github.com/arc42/arc42-template/raw/master/dist/arc42-template-${->lang}-${->help}-asciidoc.zip",
+            req42:"https://github.com/Hruschka/req42-framework/raw/main/dist/req42-framework-${->lang}-${->help}-asciidoc.zip"
+        ]
+        def languages = [
+                arc42: ['CZ', 'EN', 'DE', 'ES', 'IT', 'NL', 'UA'], 
+                req42:['EN', 'DE']
+        ]
         for (int i=1; i<20; i++) {
             def template = System.getenv('DTC_TEMPLATE' + i) ?: ""
             if (template) {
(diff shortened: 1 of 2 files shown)


scripts/downloadTemplate.gradle
```diff
@@ -30,15 +37,15 @@ task downloadTemplate(
             template = templates.keySet()[(ant.template as Integer)-1]
         }
         println "${color 'green', 'Install '+template+' documentation template.'}"
-        def targetDir = new File(new File(docDir, inputPath), 'arc42')
+        def targetDir = new File(new File(docDir, inputPath), template)
         targetDir.mkdirs()
-        if (template=='arc42') {
-            println "For more information about arc42 see https://arc42.org"
-            /**
-             if (targetDir.exists()) {println "${targetDir.canonicalPath} already exists"
```
PacoVK, line 38: you removed this, i think it was useful to let the user know that there is already a setup and the intentional re-install requires manually removal of the existing structure

scripts/downloadTemplate.gradle
```diff
@@ -30,15 +37,15 @@ task downloadTemplate(
...
-             println "in order to re-install the template, please remove the folder first and re-run the script"
-             throw new RuntimeException("Template folder already exists")}**/
-            ant.input(message: "${color 'green', 'Which language do you want to install?'}", validargs: 'CZ,EN,DE,ES,IT,NL,UA', addproperty: 'lang')
+        if (template in ['arc42', 'req42']) {
```
PacoVK, line 42: isn't that already ensured with line 36-37?

scripts/downloadTemplate.gradle
rdmueller: It was commented out. Yes, the origianl intention was to warn the user. But as averathing is under source control, I originally commented it out and now cleaned the code up

rdmueller: But if you think it is useful, we can re-enable it

scripts/downloadTemplate.gradle
```diff
@@ -30,15 +37,15 @@ task downloadTemplate(
...
+            if (template=='arc42') {
+                println "For more information about arc42 see https://arc42.org"
+            } else {
```
PacoVK, line 45:
```suggestion
            } else if (template == 'req42') {
```
We could throw an exception and abort in the default `else` here, if L36-37 does not already cover that
Alternative: use a switch case, throw an exception in the `default` block - but that just a matter of personal preference i guess

scripts/downloadTemplate.gradle
rdmueller: Line 23 allows to configure more local templates which are not part of this list. Those templates currently can't have a language or other features. Instead, you have to configure one template for each feature

rdmueller (commented):
Thanx for your review!

scripts/downloadTemplate.gradle
PacoVK: Argh, did't recognize the block comment. However, i do think it's useful to make this explicit to the user, that re-install is not possible on existing projects

scripts/downloadTemplate.gradle
PacoVK: from your comments above,  i think we can ignore my suggestion then

PacoVK (approved):
LGTM

Issue #1053, opened 2023-02-23
publishToConfluence: hash is displayed at end of some pages


rdmueller:
the hash should be hidden behind an html comment, but it is shown on some pages.

hash not visible: https://arc42-template.atlassian.net/wiki/spaces/v8sde/pages/2054553601/1.+Einf+hrung+und+Ziele

hash visible: https://arc42-template.atlassian.net/wiki/spaces/v8sde/pages/2054619137/2.+Randbedingungen

PacoVK:
solution provided in #1189 using the <ac:placeholder> tag

PacoVK:
Fixed by #1189

Issue #1186, opened 2023-05-25
Execution failed for task ':publishToConfluence'. > java.io.FileNotFoundException: /project/build/html5/../images/X.png (No such file or directory)


stefanitsky:
**Describe the bug**

Can't publish docs to confluence.

**To Reproduce**
1. Download template (https://github.com/bitsmuggler/arc42-c4-software-architecture-documentation-example);
2. Configure project (confluence creds, namings etc.);
3. Execute publish (publishToConfluence);
4. See error: 
<img width="1000" alt="image" src="https://github.com/docToolchain/docToolchain/assets/22547214/45f2ff59-2c7b-47bd-84b3-f1c5c3ad6666">

**Expected behavior**

Documentation successfully publishes.

**Configuration**
 - docToolchain version: ng
 - OS: OSX + docker (custom image based on eclipse-temurin:11-jdk-jammy)

I fixed it by changing this line: https://github.com/docToolchain/docToolchain/blob/ng/scripts/asciidoc2confluence.groovy#L653
But i don't know, is it ok or not.

Project tree:
```
├── Config.groovy
├── Dockerfile
├── Makefile
├── README.md
├── build
│   └── html5
│       ├── projectX.html
│       └── images
│           └── auth_by_login_and_password.png
├── docs
│   ├── adrs
│   │   ├── 0001-record-architecture-decisions.adoc
│   │   ├── 0001-record-architecture-decisions.md
│   │   ├── 0002-auth-with-jwt.adoc
│   │   ├── 0002-auth-with-jwt.md
│   │   ├── 0003-arc-42-docs-format.adoc
│   │   └── 0003-arc-42-docs-format.md
│   ├── projectX.adoc
│   ├── arc42
│   │   ├── 01_introduction_and_goals.adoc
│   │   ├── 02_architecture_constraints.adoc
│   │   ├── 03_system_scope_and_context.adoc
│   │   ├── 04_solution_strategy.adoc
│   │   ├── 05_building_block_view.adoc
│   │   ├── 06_runtime_view.adoc
│   │   ├── 07_deployment_view.adoc
│   │   ├── 08_concepts.adoc
│   │   ├── 09_architecture_decisions.adoc
│   │   ├── 10_quality_requirements.adoc
│   │   ├── 11_technical_risks.adoc
│   │   ├── 12_glossary.adoc
│   │   ├── _about-arc42.adoc
│   │   └── _config.adoc
│   └── diagrams
│       ├── microservice_x
[...]

PR #1189, opened 2023-05-25
Improve publish to confluence 

PacoVK:
related to #1186  embedded images are quite flaky to handle, there are too many moving parts. 
Hence, we must document the constraints when enabling `failOnMissingImages` and the default should just ignore missing images. 

This PR also adds several unit tests for `publishToConfluence` task.

**New features:**
* Confluence Editor is configurable (default: v1)

**Fixes**
* Page hash is always invisible in read mode
* Pages with new Editor are fullwidth again [reference](https://community.developer.atlassian.com/t/how-to-set-fixed-width-when-creating-new-page-with-rest-call/53591/13)


PacoVK:
Test setup keeps in line with the propsal of #1178 @mh182

PacoVK:
closes #1053

rdmueller (approved):
LGTM! Thanx - this is an important change!

rdmueller:
we could add a changelog entry, couldn't we?

rdmueller:
mh. all builds start to fail. there seems to be a problem with maven. Old version of libraries seem to be missing.

PacoVK:
I guess it more an error on a mirror because it is available on mvncentral eg. https://mvnrepository.com/artifact/com.burgstaller/okhttp-digest/1.10 I hope i magically disappears soon, that we can proceed to integrate :D

Issue #1193, opened 2023-06-09
exportEA fails since v2.0.5 with an exception in line 106


PeterStange:
**Describe the bug**
Up to tag v2.0.3 of branch 'ng' there was no issue using exportEA to export diagrams and more from an EnterpriseArchitect project. 
With updating to the next available tag v2.0.5 issues are observed in task streamingExecute of exportEA.gradle. There was a change made to cover exportPPT issues with https://github.com/docToolchain/docToolchain/issues/420 exportEA encoding problem.

We get this error after applying this change:
FAILURE: Build failed with an exception.

* Where:
Script 'D:\repositories\docToolchain\scripts\exportEA.gradle' line: 106

* What went wrong:
Execution failed for task ':exportEA'.
> java.lang.Exception:
  sorry, this task currently needs needs windows as operation system
  if you started this task from WSL, please switch to powershell

**To Reproduce**
1. Clone docToolchain tag 'v2.0.5'
2. Step into the root folder and simply call 'gradle exportEA'
3. See error

**Expected behavior**
The model 'docToolchain\src\docs\Models.eap' is found by the exportEA task and all diagrams were exported.

**Screenshots**

**Configuration**
 - docToolchain v2.0.5
 - OS: Window
 - Browser Firefox
 - Relevant SDK Versions [e.g. JDK 1.8, ...] 
 - gradle 6.7.1, openjdk version "1.8.0_372"

**Additional context**
Problem is caused by line: 
'String commandUTF8 = new String(command.value, 'utf-8');'
message: is "No such property: value for class"

locally fixed by replacing erroneous line by 
'String commandUTF8 = new String(command.toString().getBytes(), 'utf-8');'
Advantage of doing it this way: it works also in case command is already a String as String.toString() works also.

Any help is appreciated to get the right fix here. Even the reason for placing this charset conversion is not clear to me. Maybe someone can explain.
@rdmueller : what is your suggestion to get it generally fixed. We would like to switch latest available version, but with this issue it is not possible.

PacoVK:
Hey @PeterStange  could you please try v2.2.1 as this is the current latest release? Does the problem still persist?

PeterStange:
Hey @PacoVK, yes the problem exists in latest version v2.2.1, too. 
I observed the issue in latest version and analysed for last working one, which is v2.0.3.

PacoVK:
[...]

PR #1196, opened 2023-06-12
Fix issue #1193 with String constructor on converting to utf-8

PeterStange:
As described in #1193 the task exportEA fails with an exception in line 106 on latest version v2.2.1. The failing function was 'task streamingExecute'.
The message shown is not the right reason as the execution is made on Windows. The real reason for the issue is the String constructor. By using the toString().getBytes() function a byte array is created that is valid for the String constructor. It also works in case the command object is already of type String.

The fix implemented by this pull request is working with latest commit on branch 'ng', with tag v2.2.1 and older ones.

### DIFF
--- a/changelog.adoc
+++ b/changelog.adoc
@@ -9,6 +9,7 @@ and this project tries to adhere to https://semver.org/spec/v2.0.0.html[Semantic
 
 === fixes
 
+* https://github.com/docToolchain/docToolchain/issues/1193[#1193: exportEA fails since v2.0.5 with an exception]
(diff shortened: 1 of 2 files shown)


PacoVK:
Hey @PeterStange  would you mind to add the fix to the changelog please? 

```
* https://github.com/docToolchain/docToolchain/issues/1193[#1193: exportEA fails since v2.0.5 with an exception]
```

Apart from that thanks for your investigation an PR! We can integrate it.

PeterStange:
> Hey @PeterStange would you mind to add the fix to the changelog please?

Of course, sorry for missing this activity.

PacoVK (approved):
👍 thanks again

Issue #272, opened 2019-02-16
Generate antora documentation module 


c835722:
Hi. Thanks for this project and the opportunity to particpate in feature requests.

Feature objective:
docToolchain integration with Antora such that docToolchain would generate an Antora documentation module.

End game objective:
Architecture, Dev, Test projects documents housed under a **_single_** Project website. i.e. I envisage a site like https://doc.owncloud.org/server/index.html but also containing the project Architecture work products, models and diagrams.

rdmueller:
Great feature request and we are indeed already working  on such a feature.
Currently, we use jekyll to combine all docs into a full website since jekyll even has some more features than antorra. An example is https://docs-as-co.de

And Andreas works on an Antora version of arc42: https://blog.anoff.io/2019-02-15-antora-first-steps/

We'll keep you updated!

PacoVK:
What is the actual status here? I would like to take a look into this issue, but honestly need more input here :) 
Basically the overall question is if the main goal would be to integrate an Antora template or even be able to auto-generate Antora module(s) from the docs. 
For the latter i could envision to have a concept similar to github-pages where the actual result (the antora module) could be integrated into any existing Antora stack with a reference in playbook.yml to this particular branch. 
That may be validated if that would be possible.

rdmueller:
the status is that we don't have enough time to implement everything.
But if you want to add features to docToolchain, feel free to propose your own.

I never managed to use antora and tehrefore don't know about antora modules, but I could imagine two interesting features:

- let jBake render an Antora document structure
- use Antora instead of jBake

What do you think?

PacoVK:
alright, thanks for the insights :) Just wanted to ensure that nobody is already working on that, since some older comments stated that. Ofc i see time is always a big constraint. 
I will wrap my head around and come up with a propsal, i basically need to dig a bit deeper into jBake first to be able to come up with a evidence-based proposal. 
Will use this issue to keep you posted and to have some discussions on how to proceed 👍

rdmueller:
I can help you with the jbake part...

PacoVK:
I think that would be great 👍

rdmueller:
let's try to sketch some infos here. Maybe we can turn it into some docs later.

jBake first parses all pages, converts them and stores the results in an inmemory database.

It then uses the templates to render the final pages. Execute `./dtcw copyTheme` to copy the default jBakeTheme to your respository (src/site/templates).

[...]

PR #1197, opened 2023-06-16
Antora Integration

PacoVK:
## What this PR is about

It basically refers to #272 and aims for an integration with [Antora](https://antora.org/). 
The idea is that docToolchain already does the heavy lifting and boilerplate for you to setup a Antora module during the init (more specific `downloadTemplate` task). This enables you to use docToolchain in a context of Antora and integrates all advantages of both worlds, like distributes, but centralized docs with multiple output channels. 

## What is the current status of this PR?

Currently this PR is considered to be highly in progress. You can think of a PoC for the Antora Integration. 
The code needs also to be productionified, and maybe we should add tests, because there is a lot of black magic in here :D 
Feedback is appreciated and best to be added as comments here. 

## Limitations

The inital version will only be compatible with the default templates (`arc42`, `req42`), of course you can add a custom template, but you need to ensure the Antora setup is correct. 

### All Submissions:


rdmueller (approved):
looks good to me!

PacoVK:
For full compatibility to the Req42 template https://github.com/Hruschka/req42-framework/pull/3 needs to be merged

PR #1213, opened 2023-07-15
Fix document structure by adding list continuations

### LINKED ISSUE #1185: Docs: Feedback for 'publishToConfluence'
The page structure is broken because some of the list continuations are missing.

I'm also not sure if numbered lists are the best representation for those sub steps.

---
#page: https://github.com/doctoolchain/doctoolchain/edit/ng/src/docs/020_tutorial/070_publishToConfluence.adoc
#branch: ng

johthor:
And switch from markdown source blocks to listing or literal blocks.

Issue: #1185

### All Submissions:


src/docs/020_tutorial/070_publishToConfluence.adoc
```diff
@@ -26,38 +26,44 @@ In this tutorial, you'll learn how to publish the arc42 template to a Confluence
...
 
 . In the Terminal, type the following:
 +
-[source, bash]
+[source, console]
 ----
-mkdir publishToConfluenceDemo
-cd publishToConfluenceDemo
-curl -Lo dtcw doctoolchain.org/dtcw`
+$ mkdir publishToConfluenceDemo
```
rdmueller, line 36: I am not so sure about this change. The prefix makes it look more like a console, but what about copy & paste? I guess it will be broken.

We could move the $ to the css as a prefix.

What do you think?

rdmueller (approved):
great. thanx for taking care of the docs.

regarding the one comment - feel free to decide what is best.

src/docs/020_tutorial/070_publishToConfluence.adoc
johthor: I've taken a look at the [AsciiDoctor documentation](https://docs.asciidoctor.org/asciidoctor/latest/cli/) itself. They use a simple indented literal block for single commands with the `$` included. But on the rendered page this is transformed into a code listing block highlighted with highlight.js and language shell and the `$` is somehow removed from the line when copied. I'm not sure how that is achieved.

source:
```asciidoc
 $ asciidoctor --version
```

rendered page:
```html
<code class="language-console hljs shell" data-lang="console">
  <span class="hljs-meta">$</span> asciidoctor --version
</code>
```

I would prefer to have the `$` being shown, as you pointed out it makes it stand out as a command being entered. I had hoped the switch from `bash` being a shell script to `console` would already do the trick.

And after some more research this seems to be achieved via the CSS proper `user-select: none` on the `.hljs-meta` elements.

The best way forward for me seem to be to mark console commands as `[source, console]` and add some CSS properties.

johthor: And just now I realized that docToolchain does not use highlight.js for code highlighting on the microsite but google-code-prettify. Which is seems to be no longer supported https://github.com/googlearchive/code-prettify.

So now I don't know how to achieve having the `$` and not breaking the copy&paste flow.

rdmueller: ok. let's switch to highlightjs...

rdmueller: I tried to implement highlightjs in this branch and pushed the result, but I don't know where they where pushed to :-o

So I created a new issue and PR: https://github.com/docToolchain/docToolchain/pull/1238

rdmueller:
@johthor, I think we can merge this, can't we?

PR #1223, opened 2023-07-19
fix #1192 exportExcel rowspans

### LINKED ISSUE #1192: exportExcel: unnecessary Rowspans break rendering of table
**Describe the bug**

When each cell of a row is merged with the same cell on the next row, a table row like this will be produced:

```
.2+| ten .2+| eleven
```

As this seems to be correct, it will break the rendering of the table in Asciidoctor with an error

```
dropping cell because it exceeds specified number of columns
```

**To Reproduce**

Here is an AsciiDoc snippet which reproduces the Asciidoctor bug:

```
|===
| one | two
| three | four
.2+| five | six
| seven
| eight | nine
.2+| ten .2+| eleven
| next | next
| next | next
|===
```

An excel file which reproduces the behaviour has to be created.

**Expected behavior**

The whole table is rendered. I guess it would be enough to remove the rowspans

rdmueller:


scripts/exportExcel.gradle
```diff
@@ -174,6 +174,43 @@ use `gradle exportExcel` to re-export files
...
+                style.eachWithIndex { s, i ->
+                    if (s!="skip") {
+                        if (s.contains('+')) {
+                            def span = s.split('[+]')[0].split('[.]')
+                            if (span[0] != '') {
+                                removed << span[0] + '+' + s.split('[+]')[1]
+                            } else {
+                                removed << s.split('[+]')[1]
+                            }
+                            println i
```
PacoVK: the println statement should be removed or contain a userfriendly message if that is necessary

PacoVK (changes requested):

scripts/exportExcel.gradle
rdmueller: good point :-)

PacoVK (approved):

PR #1224, opened 2023-07-19
Hide the task's source code and link to GitHub

### LINKED ISSUE #1207: Hide the implementation of the tasks in the documentation
> We have consent that we need to hide the source code somehow but to keep a way to easy navigate to it.

### Discussed in https://github.com/docToolchain/docToolchain/discussions/1175

<div type='discussions-op-text'>

<sup>Originally posted by **PacoVK** May 15, 2023</sup>
As of now every task has it's own documentation page. This typically consists of the "how-to-use"part and additionally the actual implementation details. 
For some tasks these informations cause a lot of text and informations to the user. 
There was recently [the discussion](https://github.com/docToolchain/docToolchain/pull/1174#issuecomment-1547639395) if we need to have the actual implementation of a task within the docs. 
The goal of this discussion is to find out whether we want to provide this implementaiotn details as part of the docs or not. 
If we want to keep this information, we should consider how this could be improved, that it's not overwhealming users. </div>

johthor:
This change will hide the task's source code inside a collapsible block as discussed in #1175 and tracked in #1207.

Additional changes:

- Add a direct link to the embedded file at GitHub.
- Push the sections down one level.
- Remove _Add a Glance_ sections with the diagram excerpts.

The implementation is heavily inspired by a technique @ahus1 presented at a recent Meetup for kind of a [poor man's macro](https://www.ahus1.de/post/art-of-templating#_reusing_content_for_rendering).

### All Submissions:

* [x] If yes, did you update the documentation or create an issue for updating it? See #1207


rdmueller:
I just checked the copyThemes task in the preview and it looks perfect to me. Great work!

btw: when you come across these (broken) diagrams which I have put at the start of many tasks (like this https://deploy-preview-1224--dtc-docs-preview.netlify.app/images/ea/Manual/exportJiraIssues.png), it would be great if you could remove them :-)

rdmueller (approved):
I love the solution and I like how you use includes with parameters :-)

src/docs/015_tasks/03_task_copy_themes.adoc
```diff
@@ -2,21 +2,19 @@
 include::_config.adoc[]
 
 [[task-copyThemes]]
-= copyThemes
```
johthor, line 5: I've switched from section level 0 to level 1 here because IntelliJ inspections were complaining about it being only allowed in documents with type book. But after comparing it with AsciiDoctor documentation source where level 0 is the highest level inside a single page and is transformed into `h1` I am no longer sure about that change.

@rdmueller what is the expected section level here and how does it translate to the microsite generated by JBake? Or is there already some documentation regarding the appropriate section levels?

rdmueller: I always use level 1 (`==`) in my documentation files because they are only chapters. When I render for example a PDF, those chapters will be assembled to a full document which gets a level 0 (`=`) headline. Level 0 is only allowed once in a document.

rdmueller:
@johthor shall we setup a working session to implement this PR together?

src/docs/015_tasks/03_tasks.adoc
```diff
@@ -5,7 +5,7 @@ include::_config.adoc[]
 
 include::../_feedback.adoc[]
 
-== At a Glance
+=== At a Glance
```
johthor: @rdmueller this is the last _At a Glance_ section remaining. Should this one also be removed? It is the only one containing a complete tasks diagram.

rdmueller: why not? let's remove it!

johthor:
@rdmueller I've now removed all sections "At a Glance" and the corresponding images.

After a few months of maturing, the work is now complete from my point of view.

I am still not sure about the correct heading level, but that is perhaps another task and needs a little more research

rdmueller:
@PacoVK we should merge this, shouldn't we? You have my approval...

PacoVK:
@rdmueller while i agree, i will do that and fix-forward the files that have changed since then

PR #1208, opened 2023-07-05
WIP: Start a POC for separation of tools and tasks

johthor:
Refs #1076

See https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources for our inspiration.

### All Submissions:

* [ ] Did you update the `changelog.adoc`?
* [ ] Does your PR affect the documentation?
* [ ] If yes, did you update the documentation or create an issue for updating it?

The source of the documentation can be found in `/src/docs/`.

If you didn't find the time to update docs, please create an issue as reminder to do so.

PR #1226, opened 2023-07-28
Confluence APIv2 

### LINKED ISSUE #1205: Upgrade to Confluence REST API v2
In order to continue the support for Confluence Cloud we must migrate our integration to make use of REST API v2 until 01.01.2024 **latest** (see [reference](https://community.developer.atlassian.com/t/deprecating-many-confluence-v1-apis-that-have-v2-equivalents/66883)

PacoVK:
# Upgrade to Confluence APIv2

## What?
We need to upgrade to the Confluence API v2 until 01.24 latest with a lot of breaking changes. See  #1205 
We must be backwards compatible to ensure older Confluence versions can still be a target for tasks like `publishToConfluence`

In #1208  there is already a sample on how we could finally separate tools and tasks. Adhering this gives a bit more convenience and flexibility to implement the solution. Hence, this PR already is follows what #1208 suggests, but this is not a must and the implementation can be moved towards the current structure.

## The idea
Basically the idea is to introduce Confluence REST Clients that hide the actual API. The user can switch according her needs between the version that is required. 

### All Submissions:


PacoVK:
The first question to be answered is, should we stick to the current project structure or not

rdmueller:
> The first question to be answered is, should we stick to the current project structure or not

What do you mean by "project structure"?

Shall we formulate this as ADR?

rdmueller:
Please advice me how I can help best

PacoVK:
Yeah, well I think an ADR was a good hint so we and others can have a discussion. I try to manage it to come up with something 👍

build.gradle
```diff
@@ -9,6 +9,7 @@ buildscript {
                 details.because 'jcenter redirect'
                 System.out.println ">>>>>>>>> fix grolifant"
             }
+            //TODO is this really needed here?
```
rdmueller, line 12: this is a safety net which helped with when jcenter was gone. Maybe we can remove it by now, but it also does no harm

PacoVK: Yeah I remember and it makes sense I just wondered if it makes sense to go with 0.7.1 instead of force downgrade with this script. Because currently the versions file says 0.7.2

rdmueller: yeah, I am also nut sure if we are not already over this...

PacoVK: i'll keep that comment, but i will exceed the scope here

PacoVK:
Now pending: the docs need some updates

PacoVK:
@rdmueller FYI now, that i consider it ready, i will create the 3.1.0-rc1 in the next days 🏎️

rdmueller:
These are great news! We also made progress with the draw.io png support. I hope I can add it vefore you create the 3.1.0 release 😁
Shall we create the release together?

PacoVK:
sure, no pressure. Just let me know, if you are done. For the time beeing i suggest to freeze the confluence script for now. 
Or did you mean the 3.1.0 release or do you want to include it already in the 3.1.0-RC1?

PacoVK:
closes #1265

PacoVK:
closes #1254

rdmueller (approved):
LGTM

PR #1235, opened 2023-08-10
fix #1231: docker: port mapping only needed for previewSite

### LINKED ISSUE #1231: docker: port mapping only needed for previewSite
running dtcw as docker container always opens port 8042, but it is only needed for `previewSite` task

rdmueller:
### All Submissions:


PacoVK (approved):
LGTM

PR #1243, opened 2023-08-21
[DECISION] Separate core logic from Gradle

PacoVK:
As per ongoing discussion in #1076 and the poc implementations on #1208 and #1226 this is finally the first iteration of creating a new ADR to document the outcome on the topic, whether we want to separate docToolchain core logic from Gradle. 

The outcome may significantly impact docToolchains future development. 
Once we came to a conclusion how we should proceed, there may be follow-up tasks that can ensure to track a potential migration of docToolchain to incorporate this very outcome. 

While most of the discussion happend in #1076  this PR is open for fine-tuning and change requests. However, the actual conversation should be kept in #1076 to not loose context.

### DIFF
--- a/src/docs/050_ADRs/ADR-2-anotherTest.ad
+++ b/src/docs/050_ADRs/ADR-2-anotherTest.ad
@@ -1,4 +1,0 @@
-:filename: 050_ADRs/ADR-2-anotherTest.ad
-## second ADR
-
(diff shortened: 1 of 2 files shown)


PacoVK:
Starting with #1226  and releases from 3.1.x this decision is accepted.

Issue #1270, opened 2023-10-29
DTC_JAVA_HOME should supersede JAVA_HOME


ascheman:
**Describe the bug**

If JAVA_HOME is set (e.g., by sdkman), and DTC_JAVA_HOME is explicitely set or computed by DTC, the latter should be used.

**To Reproduce**
1. Install local (DTC) Java: `dtcw install java` -> results in `~/.doctoolchain/jdk`
2. Set JAVA_HOME to a different directory (or simply use [SDKMAN](https://sdkman.io)
3. Run `./dtcw local generateSite` -> Will use the Java from `${JAVA_HOME}` instead of the one installed by step one.
4. On many systems/architectures this might work perfectly without problems. On Apple Silicon this might result in a Gradle error `Can't load library: ...` (since JBake still uses an x86_64 native library).

PR #1271, opened 2023-10-29
Fix bash problems with homebrew and local Java installation

### LINKED ISSUE #1269: Homebrew bash conflict (on arm64)
**Describe the bug**
At least on arm64, if using a recent version of Homebrew, its bash (i.e., /opt/homebrew/bin/bash) is only an 'arm64' binary.
Hence, every internal call to bash will result in an arm64 architecture run of the respective executable/script.
The intended run of dtc (and the internal Java) as implemented by #1031 is nullified.
Instead, the script should explicitly call the dual architecture `/bin/bash` to ensure the execution remains in the `x86_64` universe.

**To Reproduce**
* Install Homebrew with bash on Apple Silicon
* Run `./dtcw generateSite`

Get the following error

```bash
arch: posix_spawnp: bash: Bad CPU type in executable
```

ascheman:
### All Submissions:

Closes #1269 + #1270

### DIFF
--- a/bin/doctoolchain
+++ b/bin/doctoolchain
@@ -3,6 +3,7 @@
 set -u
 set -e
 set -o pipefail
+test "${DTC_BASH_OPTS:-}" && set "${DTC_BASH_OPTS}"
 
 if [[ $# -lt 2 ]]; then
     echo '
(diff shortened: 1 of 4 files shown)


rdmueller (approved):
looks good!

rdmueller:
thanx!

PR #1301, opened 2023-11-20
support embedded svg images in publishToConfluence

### LINKED ISSUE #1300: publishToConfluence:Cannot handle embedded SVG images
This issue was raised by @faltfe (thanks!)
To keep the scope of #1206 i separated this one here: 

There was the observation in [another issue](https://github.com/docToolchain/docToolchain/issues/1206#issuecomment-1818620041) that `publishToConfluence` could not handle embdded SVG images. This can be reproduced if we set `failOnMissingImages=true` in docToolchian config. 

**To Reproduce**
1. ensure `failOnMissingImages=true` in docToolchian config (this is default)
2. add the following content to the adoc file you want to publish
```
[plantuml,03_technical-context,svg]
....
@startuml
skinparam componentStyle uml2

foo <-u-> bar

@enduml
....
```
3. run `publishToConfluence`
5. The page will be created without attachment 

**Expected behavior**
The attachment will be uploaded 

**Additional context**
If we use e.g. png instead of svg it works as expected

PacoVK:
closes #1300 

### All Submissions:

### DIFF
--- a/changelog.adoc
+++ b/changelog.adoc
@@ -9,6 +9,8 @@ and this project tries to adhere to https://semver.org/spec/v2.0.0.html[Semantic
 
 === fixes
 
+* https://github.com/docToolchain/docToolchain/issues/1300[#1300: publishToConfluence:Cannot handle embedded SVG images]
+
 === added
 
 === changed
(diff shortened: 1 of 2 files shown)


rdmueller (approved):
Is this all what was needed? Looks good to me!

PacoVK:
> Is this all what was needed? Looks good to me!

Indeed :D The svg is a exotic MIME compared to all others (few exceptions apply, but i think we can add them later if necessary) Eg png is image/png jpg image/jpg and only svg is image/svg+xml This breaks the code because we split the MIME by "/" and the file would become svg+xml (uri decoded "svg xml") which is then invalid :P

PR #1305, opened 2023-11-23
bug fix assert_java_version_supported

janni2:
bug fix assert_java_version_supported:
- trouble invoking java command with space in path on Windows especially C:\Program Files\GraalVM\graalvm-ce-java17-22.3.2\bin\java
- the checks are partly ignored while the order should be: local install, java from path, "fallback java from java home"
- version string: no need to perform "guessings" on return string, when powershell has build-in extraction of the version as object 

not further tested a.k.a. "works for me". Just as hint & suggestion since invokation of java is not touched and only that assertion is fixed, sothat the wrapper does not crash

### DIFF
--- a/dtcw.ps1
+++ b/dtcw.ps1
@@ -390,26 +390,33 @@ function download_file($url, $file) {
 
 function assert_java_version_supported() {
     # Defines the order in which Java is searched.
-    $JAVA_CMD = ""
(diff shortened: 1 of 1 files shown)


rdmueller (approved):
looks good to me. thanx!

rdmueller:
Welcome to the team as contributor!

Issue #1338, opened 2024-01-12
extend the options for the docker environment by additional environment variables and additional parameter sets.


MichaelRossner:
**Is your feature request related to a problem? Please describe.**
With #1326 it is now possible to use different docker images.
The extended docker images might need
* environment variables
* extra parameters
to make proper use of all additional tools installed in the docker image.

**Describe the solution you'd like**
Additional options would be nice to make use of the extended possibilities of the new docker image.

**Describe alternatives you've considered**
none

**Additional context**
none

MiRoCopilot:
Closed via #1339.

Thanks

PR #1339, opened 2024-01-12
Add additional options for docToolchain container

MichaelRossner:
solves #1338.

- add additional option to use a docker environment file
- add additional parameter to pass extra docker parameter.

### All Submissions:

* [y] Did you update the `changelog.adoc`?
* [y] Does your PR affect the documentation?
* [y] If yes, did you update the documentation or create an issue for updating it?

### DIFF
--- a/src/docs/010_manual/20_install.adoc
+++ b/src/docs/010_manual/20_install.adoc
@@ -94,7 +94,7 @@ TIP: In case your development team uses different operating systems, put the wra
 
 Once the docToolchain wrapper is installed in your project directory you have to decide how to install the toolchain:
 
-- <<run-dtc-in-container>> with the https://hub.docker.com/r/doctoolchain/doctoolchain[docToolchain container image].
+- <<run-dtc-in-container>> with the https://hub.docker.com/r/doctoolchain/doctoolchain[docToolchain container image] or <<run-dtc-in-costum-container>>.
 - <<install-dtc-with-dtcw>> in the users home directory `$HOME/.doctoolchain`
 - <<install-dtc-with-sdk>> a tool for managing parallel versions of multiple Software Development Kits.
 
@@ -120,6 +120,8 @@ If the container engine is installed you can <<first-command>>.
 The docToolchain wrapper script in your project directory will detect the
 container engine and pull the docToolchain image on the first invocation.
 
+[[run-dtc-in-costum-container]]
+=== Use your costum docker image
 Some might need to create their own docker image to add additional tooling or configurations, e.g. proxy settings.
 In this case you can pass the image name via parameter `image` directly after `docker`:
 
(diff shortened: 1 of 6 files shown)


src/docs/010_manual/20_install.adoc
```diff
@@ -153,6 +155,31 @@ In this case you can pass the image name via parameter `image` directly after `d
 .\dtcw.bat docker image <image_name> generateHTML
 ----
 
+[[add-environment-variables-to-docker]]
+=== Pass environment variables to docker containers
+To pass any environment variable to the docker container you can make use of an https://docs.docker.com/engine/reference/commandline/run/#env[environment file].
+The environment file must have the name `dtcw_docker_env.list` and must be located in the same folder as `dtcw`.
```
PacoVK: Good work, thanks for your efforts, really appreciate that :) 
One thing came to my mind after having a first look, i would rather prefer `dtcw_docker.env` as a filename than `dtcw_docker_env.list`
Mainly because it is an `.env` file and if you look at how other tools handle such files, it is mostly `something.env` 
What do you think to adopt your PR to follow this covention?

src/docs/010_manual/20_install.adoc
```diff
@@ -120,6 +120,8 @@ If the container engine is installed you can <<first-command>>.
 The docToolchain wrapper script in your project directory will detect the
 container engine and pull the docToolchain image on the first invocation.
 
+[[run-dtc-in-costum-container]]
```
PacoVK, line 123: good point, to add a separate heading here now 👍

rdmueller (approved):
as I can see, paco already did a deeper review than I, but from my point of view, this looks good! Thanx for the contribution.

src/docs/010_manual/20_install.adoc
MiRoCopilot: Hi @PacoVK,

I agree with you now 😄. I took the extension from an example - but have now seen that more use *.env.

If you still want to have this change - I can create another MR.

As long as it is done before the next release it will not break anyone's setup.

PacoVK: @MichaelRossner  i am not sure about @MiRoCopilot comment, but i guess you relate to each other :D 
The question i have is, why create another MR/PR if you could still alter this existing one to reflect the propsed changes? :)

MichaelRossner: Hi @PacoVK,

yes I was signed in with another account. I misread the comment of @rdmueller as already merged.

I upload a new change.

MichaelRossner: File has been renamed to the proposed: `dtcw_docker.env`.

PacoVK: Good work, thanks! :)

MichaelRossner:
Thanks for merging.

Issue #1335, opened 2024-01-10
publishToConfluence throws exception when uploading attachments in v3.2.0


corradolupo73:
**Describe the bug**
When uploading attachments, e.g., images, using the ConfluenceClient in 3.2.0 via the publicToConfluence task, the following error occurs:

----
dtcw 0.52 - ##DTCW_GIT_HASH##
docToolchain 3.2.0
OS/arch: pwsh windows x64
Available docToolchain environments: local docker
Environments with docToolchain [3.2.0]: docker
Using environment: docker
To honour the JVM settings for this build a single-use Daemon process will be forked. See https://docs.gradle.org/8.1.1/userguide/gradle_daemon.html#sec:disabling_the_daemon.
Daemon will be stopped at the end of the build 
> Task :core:compileJava NO-SOURCE
> Task :core:compileGroovy UP-TO-DATE
> Task :core:processResources NO-SOURCE
> Task :core:classes UP-TO-DATE
> Task :core:jar
> Task :generateHTML UP-TO-DATE

> Task :publishToConfluence
publish build/html5/index.html
Farm
.
pages retrieved
retrieving page with id 1671036945
found existing page: 1671036945 version 1
page hasn't changed!
1. Struttura documentazione
    image: /project/build/html5/images/diag-1d6529118cf5d3043239244bb749a8aa.png
allPages already retrieved
pages retrieved
retrieving page with id 1671036946
found existing page: 1671036946 version 1
page hasn't changed!
Uploading attachment to /rest/api/content/1671036946/child/attachment

FAILURE: Build failed with an exception.

* Where:

> Task :publishToConfluence FAILED
Script '/home/dtcuser/docToolchain/scripts/publishToConfluence.gradle' line: 47

* What went wrong:
Execution failed for task ':publishToConfluence'.
> No such property: ContentType for class: org.docToolchain.atlassian.confluence.clients.ConfluenceClientV1
----

**To Reproduce**
Install 3.1.0 docToolChain Docker image
docToolchain 3.2.0
OS/arch: pwsh windows x64
Download arc42 template using ./dtcw downloadTemplate
Run ./dtcw publishToConfluence ...
See error
1. OS/arch: pwsh windows x64
2. Install 3.2.0 docToolChain Docker image
3. Referer on section a plantuml diagram:

[plantuml, format=png]   
....
include::plantuml/documentazione.puml[]
....

**Expected behavior**

**Screenshots**

**Configuration**
[...]

PR #1340, opened 2024-01-12
Improve basic rest client

PacoVK:
As an outcome of the discussion in #1335 there are further improvements to the rest client, namely:
* Ensure the `Host` header does not contain port definitions
* Set custom `User-Agent` to ease network traffic inspections, the User-Agent will be in format `docToolchain-v<dtcw_version>`

### All Submissions:

### DIFF
--- a/core/src/main/groovy/org/docToolchain/http/BasicRestClient.groovy
+++ b/core/src/main/groovy/org/docToolchain/http/BasicRestClient.groovy
@@ -16,7 +16,7 @@ abstract class BasicRestClient {
     BasicRestClient() {
         this.httpClientBuilder = HttpClientBuilder.create()
         httpClientBuilder.addRequestInterceptorFirst { request, entityDetails, context ->
-            request.setHeader('User-Agent', "docToolchain_v${getClass().getPackage().getImplementationVersion()}")
+            request.setHeader(HttpHeaders.USER_AGENT, "docToolchain_v${getClass().getPackage().getImplementationVersion()}")
             Header hostHeader = request.getHeader(HttpHeaders.HOST)
             if (hostHeader == null) {
                 String host = new URIBuilder(request.getUri().toString()).getHost()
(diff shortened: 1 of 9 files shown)


rdmueller (approved):
great work. thanx!

core/src/main/groovy/org/docToolchain/http/BasicRestClient.groovy
```diff
@@ -3,15 +3,27 @@ package org.docToolchain.http
...
 abstract class BasicRestClient {
 
     protected HttpClientBuilder httpClientBuilder
 
     BasicRestClient() {
         this.httpClientBuilder = HttpClientBuilder.create()
+        httpClientBuilder.addRequestInterceptorFirst { request, entityDetails, context ->
+            request.setHeader(HttpHeaders.USER_AGENT, "docToolchain_v${getClass().getPackage().getImplementationVersion()}")
+            Header hostHeader = request.getHeader(HttpHeaders.HOST)
+            if (hostHeader == null) {
```
maksymgendin, line 21: @PacoVK Are you sure that this will fix the discussed issue with the port in the `Host` header? Maybe we should in all cases (not only when it's `null`) set the header manually? Not sure.

PacoVK: Unless you do not set it explicitly, i am sure at least the docs say for `addRequestInterceptorFirst` "Adds this protocol interceptor to the head of the protocol processing list." ([see](https://hc.apache.org/httpcomponents-client-5.2.x/current/httpclient5/apidocs/org/apache/hc/client5/http/impl/classic/HttpClientBuilder.html)) whilst the standard `RequestTargetHost` Interceptor later in the chain first checks, whether a `Host` header has already been set. ([see](https://github.com/apache/httpcomponents-core/blob/master/httpcore5/src/main/java/org/apache/hc/core5/http/protocol/RequestTargetHost.java)) :)

PacoVK: Although i.am not sure if the check is necessary and we would have ever the need to set custom `Host` header, this at least has then the same behaviour as the standard interceptor

PR #1347, opened 2024-01-20
Update 03_tasks.adoc

rdmueller:
Re-Added overview diagram

I often use this diagram to show how much functionality is included in docToolchain. So let's re-add it.


PacoVK (approved):
For sure a good idea :)

PR #1354, opened 2024-01-28
Fixes using a wrong arch(itecture) on macOS (Darwin)

ascheman:
Fixes using a wrong `arch` on macOS (Darwin) + Enforce x86 Java as long as JBake does not support arm64 processors

- If you happen to have another `arch` binary or script on `${PATH}` this was used instead of /usr/bin/arch by DTCW - with disturbing results. 
The patch ensures the right executable is used to perform docToolchain on macOS.
- DTCW uses x86_64 arch binaries (e.g., bash) on macOS with arm64 processor. 
Consequently, it must also enforce installation of x86 Java for local usage.

### All Submissions:

### DIFF
--- a/changelog.adoc
+++ b/changelog.adoc
@@ -9,6 +9,7 @@ and this project tries to adhere to https://semver.org/spec/v2.0.0.html[Semantic
 
 === fixes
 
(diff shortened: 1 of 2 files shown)


rdmueller (approved):
looks good to me

rdmueller:
thanx for providing this change!

Issue #1363, opened 2024-02-09
Document usage of custom Ruby extensions 


PacoVK:
There is an inline code comment that says we should document the usage for the setting `rubyExtensions` 
Let's do it. Docs will affect all Asciidoctor Tasks (generateDeck, generateHtml, generatePdf and generateDocbook)

Issue #1353, opened 2024-01-26
docker: Problem with 3.x images and dependencies?


rdmueller:
somehow, the 3.x docker images are slower than the 2.x ones when used in a closed environment pipeline.

the image uses some tricks to make itself selfcontained: https://github.com/docToolchain/docker-image/blob/ng-beta/alpine/Dockerfile#L45

but this seems not to work anymore. I get the feeling that the image runs in timeouts because it tries to download dependencies but has no connection to the internet from within the pipeline.

PacoVK:
Could you please check if the issue is still there in 3.3.1? :)

rdmueller:
3.3.1 still needs external dependencies. This is how I test it:

run the image:

    docker run -it doctoolchain/doctoolchain:v3.3.1

then execute the following commands

    curl -Lo dtcw https://doctoolchain.org/dtcw 
    chmod +x dtcw

turn off wifi and execute

    ./dtcw local generateSite --info

this returns the following error

    * Where:
    Build file '/home/dtcuser/docToolchain/build.gradle' line: 26

    * What went wrong:
    Plugin [id: 'org.aim42.htmlSanityCheck', version: '1.1.6'] was not found in any of the following sources:

    - Gradle Core Plugins (plugin is not in 'org.gradle' namespace)
    - Plugin Repositories (could not resolve plugin artifact 
    'org.aim42.htmlSanityCheck:org.aim42.htmlSanityCheck.gradle.plugin:1.1.6')
     Searched in the following repositories:
     Gradle Central Plugin Repository

PacoVK:
Hey @rdmueller  i think the issue is the usage of the doctoolchain image rather than a bug. I can reproduce the behaviour using the steps you described, but i can also proof this is not a bug when using the image as intended (maybe we lack some documentation here) 
Let me shed some light in here:
* using the image with the steps you described and downloading `dtcw` (again) but in the local project folder, let's the wrapper set the default runtime and gradle cache directory (`/home/dtcuser/.doctoolchain/.gradle/`) This directory is empty, because all the dependencies have been installed in another local (global) Gradle cache dir `/home/dtcuser/.gradle/`. This is exactly why you see this error. In fact we enforce another cache dir in the wrapper that is different from what we set when building the image and predownloading all dependencies ([see](https://github.com/docToolchain/docToolchain/blob/ng/dtcw#L642)) --> i don't know yet, whether this may just be some old stuff that we should get rid of, tbh

[...]

Issue #1239, opened 2023-08-14
Execute a custom asciidoctor converter for generateHTML


davbader:
**Background and description**
I am looking to start using docToolchain for our docs. 
Our documentation often includes mathematical expressions.
I use STEM blocks to write using latex syntax:
``` 
:stem: latexmath

Inline math stem:[\alpha + \beta] followed by a block equation:

[stem]
++++
\alpha + \beta = \gamma
++++
```
When generating HTML and publish to confluence, asciidoctor will generate code with the default mathjax deliiters, `\[`, `\]`, `\(`, `\)`.

However our Confluence MathJax plugin does not support the delimiters above. Thus, I followed [this tutorial](https://docs.asciidoctor.org/asciidoctor/latest/convert/custom/) to write a custom converter that changes the delimiters.

``` 
Asciidoctor::Converter.for 'html5'
# => Asciidoctor::Converter::Html5Converter

class MyHtml5Converter < (Asciidoctor::Converter.for 'html5')
    register_for 'html5'

    def convert_stem node
        style = node.style.to_sym
        Asciidoctor::BLOCK_MATH_DELIMITERS[style] = ['(mathjax-block(', ')mathjax-block)']
        Asciidoctor::INLINE_MATH_DELIMITERS[style] = ['(mathjax-inline(', ')mathjax-inline)']
        super
    end
end
```

I can now generate the HTML site using asciidoctor with:

```
asciidoctor -r ./my-html5-converter.rb doc.adoc
```
Delimiters are changed and everything renders correctly.

However, I do not know how to execute the converter using the doctoolchain.

**My preferred Solution(s)**
1. The converter is not needed and there is an easy way to change the delimiter for the `generatedHTML` and `publishToConfluence` task
2. I use the converter above and somheow configure doctoolchain to execute `my-html5-converter.rb` while running `generatedHTML`

rdmueller:
interesting problem. I am not sure if this is the solution, but you can configure doctoolchain to use ruby plugins:

https://github.com/docToolchain/docToolchain/blob/ng/scripts/AsciiDocBasics.gradle#L240

https://github.com/docToolchain/docToolchain/blob/ng/template_config/Config.groovy#L169

Does this help?

davbader:
Thank you for the reply and the suggestions.This does look promising.

I am using the default `docToolchainConfig.groovy` which was generated automatically.
I have added the extension  at top level:
```
rubyExtensions = ['my-html5-converter.rb']
```

[...]

PR #1364, opened 2024-02-09
fix loading of custom ruby extensions

### LINKED ISSUE #1290: Custom ruby extension for PDF generation
I try to add a custom ruby extension to process a document in order to modify the title page of a PDF document.

I have added 
```
  rubyExtensions = ['asciidoctor-pdf-extensions.rb']
```
in docToolchainConfig.groovy .

With `docToolchain-3.0.2` I get

```

* Where:
Script '/Users/kneissl/.sdkman/candidates/doctoolchain/3.0.2/scripts/AsciiDocBasics.gradle' line: 246

* What went wrong:
A problem occurred evaluating script.
> Cannot set the value of read-only property 'requires' for task ':asciidoctor' of type org.asciidoctor.gradle.jvm.AsciidoctorTask.


```

This is very close to the issue posted by @davidschottel in https://github.com/docToolchain/docToolchain/issues/1239#issuecomment-1714126736_

I am really new to docToolchain so I am not yet able to apply his solution to my problem.

PacoVK:
This will fix  #1290 #1239 and also complete the performance issue described in #1353 

This PR moves the setting `rubyExtensions` to the top level and makes it no longer an option within the `microsites` config. The reason is that it does not relate to `microsite`tasks since they depend on JBake and JBake currently does not support custom Ruby extensions.

A follow up to this task is #1363 where we should document the config option

### All Submissions:

### DIFF
--- a/changelog.adoc
+++ b/changelog.adoc
@@ -12,6 +12,8 @@ and this project tries to adhere to https://semver.org/spec/v2.0.0.html[Semantic
 * [Noticket] Enforce x86 Java on macOS as long as jbake does not support arm64 processors
 * [Noticket] Avoid usage of an arbitrary `arch` binary/script on `+${PATH}+` instead of the desired `/usr/bin/arch` on macOS.
 * https://github.com/docToolchain/docToolchain/issues/1353[#1353: docker: Problem with 3.x images and dependencies?]
+* https://github.com/docToolchain/docToolchain/issues/1239[#1239: Execute a custom asciidoctor converter for generateHTML]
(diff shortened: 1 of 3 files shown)


PacoVK:
@rdmueller PTAL if you have time :)

rdmueller:
Shouldn't we keep the mechanism that the gems are automatically activated when we use reveal?

PacoVK:
@rdmueller thats what this PR does  activate Ruby gems **only** when actually needed. Otherwise you see dramatic loss of performance in every docToolchain task due to the start of the required Ruby Gem proxy if defined globally...

rdmueller (approved):
LGTM

rdmueller:
let's merge...

PR #1379, opened 2024-03-01
Fix Confluence APIv2 Cursor Logic

bonzani:
The Confluence API v2 always returns a `_links` object containing the base path. This PR adds defensive code checking for existence of the `next` key in the `_links` map.


bonzani:
@rdmueller @PacoVK cool improvement moving the core functionality into it's own module!

This fix is probably something you'd like to get into the next release, because currently the `publishToConfluence` task is not working at all for API v2

PacoVK:
Thanks for the feedback and for the contribution! :)

PR #1385, opened 2024-03-19
Fix dependencies

rdmueller:
small changes which seem to make the dependency resolution for offline docker usage better


rdmueller:
see also https://github.com/docToolchain/docToolchain/issues/1353

PacoVK:
@rdmueller  i can see a lot of changes, but in the end only a small diff, which seems to be fine. We can go ahead i guess :) 
.

PacoVK (approved):

rdmueller:
the many commits are a result of my try and error development. 
It seems that the problem is party solved with an

    export DTC_HEADLESS=true

but I guess I have to invest some more time

PR #1393, opened 2024-04-12
Update asciidoc2confluence.groovy

MichaelRossner:
remove double // from trace for published pages

### All Submissions:

* [n] Did you update the `changelog.adoc`?


PacoVK:
@MichaelRossner thanks again, this is a known "bug" :) While i like this change, it may introduce the same "bug" for others depending how they set their API URL. Furthermore the output also depends on the Confluence Client used (V1 has another api path than V2, which needs to be removed). 
To fix this we probably need to introduce something like a `getConfluenceWebUrl()` on the ConfluenceClient level and use this for the log.

MichaelRossner:
@PacoVK: Thanks for the info - I was not aware of the complexity here.

Let me check if I can bring up a corresponding MR for the issue at hand.

PacoVK:
@MichaelRossner PR would be awesome if you feel comfortable with it. Indeed this is one of the open things that came along with the heavy refactoring of the task. However, since then we have many improvements and a lot of tests that make contributors life way easier and changes less risky. 
If you need any help or feedback please feel free to ping 😊

PR #1397, opened 2024-05-04
Move underlying code of publishToConfluence into class

seflue:
### What changed

The main part of **publishToConfluence** was executed in a Groovy script, which was evaluated from the Gradle task.

To increase developability of the feature, move the code from the script into an actual Groovy class and call it from the Gradle task.

### Motivation
I create this as a draft PR. If you relate with the idea, please give me a hint. I then would actually use the opportunity to make the new class a bit more testable and add some tests.

Because I didn't found any tests for publishToConfluence, I just tested it out with a testspace in the Confluence Cloud instance I have access to. How do you actually tested this in the past? Is there a collection of fixtures I missed?

The motivation behind this PR is, that I am going to use DocToolChain quite often in the upcoming future (especially the "publishToConfluence" feature). I learned about it in a workshop @rdmueller gave last year and introduced it in my current project. Because I love to improve the tools I am using, I would like to contribute to DocToolChain.

### DIFF
--- a/scripts/publishToConfluence.gradle
+++ b/scripts/publishToConfluence.gradle
@@ -19,6 +19,8 @@ buildscript {
 
 import org.docToolchain.tasks.VerifyConfluenceApiAccessTask
 import org.docToolchain.tasks.WipeConfluenceSpaceTask
+import org.docToolchain.tasks.Asciidoc2ConfluenceTask
+import org.docToolchain.scripts.*
 
 //tag::verifyConfluenceApiAccess[]
 task verifyConfluenceApiAccess(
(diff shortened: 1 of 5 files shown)


rdmueller (approved):
Looks good to me, but I guess @PacoVK should also have a look at it.

rdmueller:
thanx for this contribution!

seflue:
@rdmueller If you don't mind, you can merge now. When I have time to write some tests, I will create another PR. But for me, the changes just work fine.

PacoVK:
Thanks for the good work! I will take a look and then merge it. Apparently I waited because of the draft mode so far 😊🖖

scripts/publishToConfluence.gradle
```diff
@@ -68,9 +70,8 @@ task publishToConfluence(
         config.confluence.useV1Api = findProperty("confluence.useV1Api") != null ?
             findProperty("confluence.useV1Api") : config.confluence.useV1Api != [:] ?
             config.confluence.useV1Api :true
-        binding.setProperty('config',config)
-        binding.setProperty('docDir',docDir)
-        evaluate(new File(projectDir, 'core/src/main/groovy/org/docToolchain/scripts/asciidoc2confluence.groovy'))
+        println("Using dev version!")
```
PacoVK: Please delete this single debug print ✌️

PacoVK (changes requested):
One remark and then we are good to go, good job👏 
Please feel free to put you into the contributors file as well

scripts/publishToConfluence.gradle
seflue: Done. :+1:

PacoVK (approved):

PR #1404, opened 2024-06-04
Fix typo in install instructions

r0ckarong:
I updated the docs to remove a typo for an ID and the xref use of it. No changelog entry needed.


rdmueller:
thanx!

PR #1422, opened 2024-07-23
fix jira inline macro

### LINKED ISSUE #1411: Jira links substitution is not working after migrating from 2.1.0 to 3.3.1
**Describe the bug**
After migrating from v2.1.0 to v3.3.1 I noticed that the Jira link substitution stopped working.

**Expected behavior**
If we use for example `jira:ARC-102[]` in our adoc src file, when we run the task `generateHTML` the resulting HTML files should have the above part replaced with the Jira link pointing to our Jira server.

**Configuration**
 - docToolchain version: 3.3.1 Docker image

PacoVK:
closes #1411 

### All Submissions:

### DIFF
--- a/core/src/main/groovy/org/docToolchain/scripts/asciidoc2confluence.groovy
+++ b/core/src/main/groovy/org/docToolchain/scripts/asciidoc2confluence.groovy
@@ -1,5 +1,10 @@
 package org.docToolchain.scripts
 
+/**
+ * THIS SCRIPT HAS BEEN DEPRECATED. IT IS NOT USED ANYMORE. PLEASE REFER TO THE NEW Asciidoc2ConfluenceTask
+ * IMPLEMENTATION INSTEAD. REFERENCE ONLY.
+ */
+
 import org.docToolchain.atlassian.transformer.HtmlTransformer
 
 /**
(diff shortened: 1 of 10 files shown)


core/src/main/groovy/org/docToolchain/scripts/asciidoc2confluence.groovy
```diff
@@ -1,5 +1,10 @@
 package org.docToolchain.scripts
 
+/**
+ * THIS SCRIPT HAS BEEN DEPRECATED. IT IS NOT USED ANYMORE. PLEASE REFER TO THE NEW Asciidoc2ConfluenceTask
+ * IMPLEMENTATION INSTEAD. REFERENCE ONLY.
+ */
+
```
rdmueller, line 7: good to know 😁

rdmueller (approved):
LGTM

rdmueller:
wow. that was quite a change!

PacoVK:
@rdmueller following the boy scout rule the change was bigger than it would have been to fix only, but the docs were also incomplete and i think i reduced a bit of complexity removing at least one ambigious config parameter that was only used in that area :)

Issue #1419, opened 2024-07-19
Tests fail for JDK 21


rdmueller:
https://github.com/docToolchain/docToolchain/actions/runs/10004519491/job/27653356357

> Task :core:compileGroovy FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':core:compileGroovy'.
> BUG! exception in phase 'semantic analysis' in source unit '/home/runner/work/docToolchain/docToolchain/core/src/main/groovy/org/docToolchain/atlassian/confluence/ConfluenceService.groovy' Unsupported class file major version 65

@PacoVK - do you already have an idea?

PacoVK:
We use Gradle 8.1.1 but Java 21 is supported since [8.4](https://docs.gradle.org/8.4/release-notes.html) I started the upgrade on #1298 but this is a tricky one, and i am not sure if we can make due to dependency issues

rdmueller:
ok, so if this is gradle releated, let's stop all efforts to upgrade

PR #1423, opened 2024-07-23
remove Java 21 from build

PacoVK:
We cannot support Java 21 as of now. See #1419

### DIFF
--- a/.github/workflows/default-build.yml
+++ b/.github/workflows/default-build.yml
@@ -30,7 +30,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         # TODO: Will it really run on one of these? , windows-latest, macOS-latest]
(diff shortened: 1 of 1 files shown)


rdmueller:
hm. shall we remove it or let it stay to remind us?

PacoVK:
i am not sure how fast we can make it to 21 because there may be heavy rewrites to get rid of some dependencies that block us (remind, we could upgrade to 8 only through workaround for same reason and i fear this time it gets even harder) 
So i would vote to remove it, because red CI steps cause noice and emails for the sake of reminding that we are not able to upgrade :D An issue would be the better choice i guess

rdmueller (approved):
ok

rdmueller:
ok. btw: I already suceeded with a basic replacement of jbake. For the next version, I want to workaround gradle

PR #1425, opened 2024-07-25
dirty fix for Jira integration

PacoVK:
This is a dirty fix for #1422  

If i find more time i will make it clean, otherwise i will do that separately.


rdmueller (approved):
LGTM

PR #1428, opened 2024-08-13
Update HTML Sanity Check to 2.0.0-rc0 and fix bugs found by it

ascheman:
### All Submissions:

@PacoVK or @rdmueller I'd be happy if you could soonish release a 3.4.1 (as this is mostly a bugfix release) which would help to build HSC itself in a clean way. 

Background: HSC makes use of the JBake templates delivered with DTC. If they contain errors, HSC cannot be built in a clean way.

### DIFF
--- a/changelog.adoc
+++ b/changelog.adoc
@@ -16,6 +16,7 @@ and this project tries to adhere to https://semver.org/spec/v2.0.0.html[Semantic
 
 === changed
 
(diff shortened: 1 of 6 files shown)


PacoVK (approved):
LGTM! I like that HSC complaints about some a11y findings ❤️

PR #1433, opened 2024-08-15
fixing execution of rubyExtension with relative path

marcozwyssig:
Summary

This update enables the use of relative paths in the Ruby extension when the rootPath and fullPath share the same root. A relative path is used if they share the same root; otherwise, the full path is retained.

This improves path handling by using relative paths when possible.


PacoVK (approved):
LGTM, thanks for your contribution! :)

rdmueller:
Tank you!

PR #1473, opened 2025-01-17
exportEA: Add option to include GUID into diagram names

MichaelRossner:
To guarantee that no diagram is overwritten at export the GUID can be added.

So if multipe groups in different packages create a diagram with the same name - both will be correctly available after export.

### All Submissions:

* [not yet] Did you update the `changelog.adoc`?
* [yes] Does your PR affect the documentation?
* [yes] If yes, did you update the documentation or create an issue for updating it?


rdmueller (approved):
looks good

rdmueller:
thanx for this PR!

PR #1584, opened 2026-06-19
fix: use CompositeConfiguration for v4 jBake site generation

raifdmueller:
## Summary
- **Fixes empty v4 docs**: Writing microsite config to `jbake.properties` flattened `Map`/`List` values to strings, causing 172 of 185 GSP template renders to fail with `NullPointerException`
- Uses `CompositeConfiguration` to pass config directly to jBake's `Oven`, preserving rich object types (Maps, Lists) that GSP templates need
- Adds null guards for `config.site_menu` in `submenu.gsp` and defaults `site.menu` to `[:]` when not configured

## Root cause
jBake's `DelegatingTemplateEngine` rebuilds the template config HashMap on every render (including `<%include%>` calls), translating dot-separated keys to underscore keys. When Maps were flattened to strings in `jbake.properties`, `config.site_menu` became a String instead of a Map, causing closure/iteration errors. And when `microsite.menu` was absent from project config, `config.site_menu` was null, causing NPE.

## Test plan
- [x] Local `generateSite` now bakes all 185/185 pages (was 13/185)
- [ ] CI build on this branch
- [ ] Verify deployed docs at doctoolchain.org/docToolchain/v4.0.x/

🤖 Generated with [Claude Code](https://claude.com/claude-code)


raifdmueller:
Fix already merged via #1583 (commits were stacked on the same branch).

PR #1690, opened 2026-08-09
dtcw: Docker Compose and devcontainer as native environments

### LINKED ISSUE #1689: dtcw: support Docker Compose and devcontainer as native environments
**Is your feature request related to a problem? Please describe.**

`dtcw` knows the `local`, `sdk` and `docker` environments. Two setups that are common in practice are missing:

* **Docker Compose** — the project runs docToolchain as a long-lived compose service. Invoking a task means typing `docker compose exec doctoolchain bash -c 'doctoolchain /project generatePDF $DTC_OPTS'` every time.
* **Devcontainer** — VS Code / `devcontainer` CLI setups where docToolchain sits in a sidecar container declared in `devcontainer.json`.

In both cases a container with docToolchain is already running and ready, yet `./dtcw generatePDF` cannot use it and falls back to a host installation (or a cold `docker run`).

Add `compose` and `devcontainer` as first-class environments, auto-detected ahead of the host installations — a running container needs no install step and no cold start:

```
compose → devcontainer → local → sdk → docker
```

* `compose`: a `docker-compose.yml` with a service whose image matches `doctoolchain/` (overridable via `DTC_COMPOSE` / `DTC_SERVICE`), plus a working compose command (`docker compose`, `docker-compose`, `podman-compose`).
* `devcontainer`: `.devcontainer/devcontainer.json` referencing a doctoolchain image, plus the `devcontainer` CLI.
* Forcing an environment stays possible: `./dtcw local generateHTML`.
* `./dtcw compose install doctoolchain` and `./dtcw devcontainer install doctoolchain` bootstrap the respective config when absent.
* Container runtime detection prefers lightweight/daemonless runtimes (`container`, `nerdctl`, `finch`, `podman`) before falling back to `docker`.

Dispatch reuses the existing `build_command()`/`exec` path via a command prefix, so there is no second invocation route to keep in sync. Inside the container, v3 calls `doctoolchain` while v4 goes through `Launcher.groovy` (ADR-18) — task discovery is therefore delegated to the container instead of being answered on the host, which has no view of what the image provides.

**Describe alternatives you've considered**

A sourceable shell function (`lib/dtcw-function.bash`) with multi-mode dispatch. Rejected: the script degenerates into a passthrough to the function, and users have to source files in their shell profile. The same pattern was explored in `tpo42-asciidoc-container`, whose ADR-007 reached the same conclusion.

**Additional context**

Rationale and the environment/dispatch design are captured as ADR-19 in the arc42 register.

Two pre-existing `dtcw` defects surfaced while building this. They are unrelated to the feature and are kept as separate, self-contained commits:

1. `docker compose exec` allocates a pseudo-TTY by default, so the compose path aborts with *"the input device is not a TTY"* whenever stdin is not a terminal (CI, pipes, cron). The wrapper already derives `DTC_HEADLESS` from `[ -t 0 ]` — it just was not consulted.
2. `build_command()` returns a string that `main()` hands to `bash -c`, so that shell re-parses it. `-PmainConfigFile=${DTC_CONFIG_FILE}` and `${dtc_home}/bin/doctoolchain` were interpolated unquoted and get word-split on paths containing spaces. The local v3 environment therefore fails on Windows/cygwin, where `DTC_ROOT` lands under `C:\Users\First Last`. The surrounding code already uses the single-quote idiom for `-Dgradle.user.home=...`; these two places were missed.

Happy to split those two out into their own issue and PR if you would rather keep them separate.

rehsack:
Fixes #1689

Adds `compose` and `devcontainer` as native `dtcw` environments, so `./dtcw generateHTML` uses an already-running docToolchain container instead of falling back to a host installation or a cold `docker run`.

### Environments

Auto-detection order — project-local contexts that carry their own docToolchain win, because a running container needs neither an install step nor a cold start:

```
compose → devcontainer → local → sdk → docker
```

Forcing still works (`./dtcw local generateHTML`), and `./dtcw compose install doctoolchain` / `./dtcw devcontainer install doctoolchain` bootstrap the config when it is missing.

Dispatch reuses the existing `build_command()`/`exec` path through a command prefix, so there is no second invocation route to maintain. Inside the container v3 calls `doctoolchain`, v4 goes through `Launcher.groovy` with `dtc.scriptsHome` (ADR-18/ADR-17). Consequently `tasks`, task resolution and unknown-task guidance are delegated to the container — the host cannot know what the image provides.

Runtime detection prefers lightweight/daemonless runtimes (`container`, `nerdctl`, `finch`, `podman`) before `docker`.

### Commits

| Commit | Content |
|---|---|
| `Ignore editor swap and backup files` | `.gitignore` |
| `Add compose and devcontainer as native dtcw environments` | the feature, ADR-19, completions, unit tests, fixtures |
| `Make Gradle warning mode configurable via DTC_WARN_MODE` | `--warning-mode` was hard-wired to `none`; `DTC_WARN_MODE=all` surfaces Gradle deprecations during v3→v4 migration (v3 only) |
| `fix(dtcw): disable the compose-exec TTY when headless` | pre-existing defect, see below |
| `fix(dtcw): quote paths interpolated into the v3 command string` | pre-existing defect, see below |
| `style(arc42): collapse the triple blank line in chapter 9` | `asciidoc-linter` finding, surfaced by touching the chapter |

### Two pre-existing defects fixed separately

Both are unrelated to the feature and kept as their own commits so they can be cherry-picked or dropped:

1. **compose-exec TTY** — `docker compose exec` allocates a pseudo-TTY by default and aborts with *"the input device is not a TTY"* as soon as stdin is not a terminal (CI, pipes, cron). `DTC_HEADLESS` was already derived from `[ -t 0 ]`; the compose path just never consulted it. Verified against `docker compose` and `docker-compose`; `podman-compose` documents `-T` as well but was not available to test.
2. **Unquoted interpolation** — `build_command()` returns a string that `main()` hands to `bash -c`, so that shell re-parses it. `-PmainConfigFile=${DTC_CONFIG_FILE}` and `${dtc_home}/bin/doctoolchain` were unquoted and word-split on paths containing spaces, so the local v3 environment fails on Windows/cygwin where `DTC_ROOT` lands under `C:\Users\First Last`. The file already uses the single-quote idiom for `-Dgradle.user.home=...` and `-Dorg.gradle.java.home=...`; these two spots were missed.

### Testing

* `test/dtcw-unit.bash` — new, sources `dtcw` behind a `BASH_SOURCE` guard and covers compose/devcontainer/runtime detection and environment validation: 13 tests.
* BATS suite: 60 of 61 pass. `sdk_installation.bats::specific doctoolchain version not installed` fails, but **identically on a pristine `main-4.x` worktree** — it is host-dependent (the test's `PATH` does not shield a real `docker` or a real `~/.doctoolchain`), not a regression from this branch.
* `bash -n` and `shellcheck` clean on `dtcw` and `lib/completions/dtcw.bash`.
* All `.pre-commit-config.yaml` hooks pass across the full branch diff.
* CI workflow gains a separate unit-test step ahead of the BATS step.

### All Submissions

* [ ] Did you update the `changelog.adoc`? — **no.** `changelog.adoc` has not been touched anywhere in the v4 range (~200 commits since `1b4b9228`) and its `Unreleased` section still lists v3-era issues, so I followed the current practice rather than reviving the file unilaterally. Happy to add an entry if you want it maintained again.
* [x] Does your PR affect the documentation? — yes.
* [x] If yes, did you update the documentation or create an issue for updating it? — ADR-19 added to the arc42 register (`src/docs/arc42/adrs/`), wired into chapter 9's ADR index and include list.

### Note on ADR-19 format

ADR-19 follows the Status/Context/Decision/Consequences prose structure of the ADR it grew out of, not the bullet-plus-Pugh-matrix shape of ADR-16…18. Say the word and I will convert it, marking cells that need your judgement with `?` per the register's convention.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.claude/skills/dtcw/SKILL.md
+++ b/.claude/skills/dtcw/SKILL.md
@@ -1,0 +1,225 @@
+---
+name: dtcw
+description: "docToolchain Wrapper — generate HTML/PDF, publish to Confluence, export from Jira, and manage docs-as-code workflows via docToolchain."
+user-invocable: true
+---
+
+# dtcw — docToolchain Wrapper
+
+Use the `dtcw` command to generate documentation, publish to Confluence, export from Jira, and manage docs-as-code workflows via docToolchain.
+
+## Prerequisites
+
+The `dtcw` wrapper auto-detects the execution environment (in priority order):
+
+1. **compose**: `docker-compose.yml` with doctoolchain service + compose command
+2. **devcontainer**: `.devcontainer/devcontainer.json` with doctoolchain image + CLI
+3. **local**: docToolchain installed in `$HOME/.doctoolchain`
+4. **sdk**: docToolchain installed via SDKMAN!
+5. **docker**: one-shot container via `docker run`
+
+Force a specific environment: `./dtcw <environment> <task>`
+
+## Tasks
+
+### Generate Documentation
+
+```bash
+# Generate HTML output
+dtcw generateHTML
+
+# Generate PDF output
+dtcw generatePDF
+
+# Generate microsite with jBake
+dtcw generateSite
+
+# Preview microsite locally
+dtcw previewSite
+
+# Generate DocBook XML
+dtcw generateDocBook
+
+# Generate reveal.js slide deck
+dtcw generateDeck
+```
+
+### Publish and Integrate
+
+```bash
+# Publish documentation to Confluence
+dtcw publishToConfluence
+
+# Export Jira issues to AsciiDoc
+dtcw exportJiraIssues
+
+# Export Jira sprint changelog
+dtcw exportJiraSprintChangelog
+
+# Export Confluence pages to AsciiDoc
+dtcw exportConfluence
+```
+
+### Export from External Tools
+
+```bash
+# Export from Enterprise Architect
+dtcw exportEA
+
+# Export Excel tables to AsciiDoc
+dtcw exportExcel
+
+# Export PowerPoint slides as images
+dtcw exportPPT
+
+# Export Visio diagrams
+dtcw exportVisio
+
+# Export draw.io diagrams
+dtcw exportDrawIo
+
+# Export Structurizr architecture diagrams
+dtcw exportStructurizr
+
+# Export OpenAPI spec
+dtcw exportOpenApi
+
+# Export Markdown to AsciiDoc
+dtcw exportMarkdown
+```
+
+### Utilities
+
+```bash
+# Export git changelog
+dtcw exportChangeLog
+
+# Export git contributors
+dtcw exportContributors
+
+# Collect all includes into a single list
+dtcw collectIncludes
+
+# Convert to DOCX via Pandoc
+dtcw convertToDocx
+
+# Convert to EPUB
+dtcw convertToEpub
+
+# Fix file encoding issues
+dtcw fixEncoding
+
+# HTML sanity check (broken links etc.)
+dtcw htmlSanityCheck
+
+# Download arc42/req42 template
+dtcw downloadTemplate
+
+# Copy site themes
+dtcw copyThemes
+
+# List all available tasks
+dtcw tasks
+```
+
+### Installation (local/project-home mode only)
+
+```bash
+# Install docToolchain locally
+dtcw install doctoolchain
+
+# Install Java 17 locally
+dtcw install java
+```
+
+## Usage Patterns
+
+### Before generating documentation
+
+When the user asks to build or preview documentation:
+
+```bash
+# Check what tasks are available
+dtcw tasks
+
+# Generate HTML for quick preview
+dtcw generateHTML
+# Output: build/html5/
+
+# Generate PDF for distribution
+dtcw generatePDF
+# Output: build/pdf/
+```
+
+### Working with Confluence
+
+```bash
+# Publish current docs to Confluence
+dtcw publishToConfluence
+
+# Export existing Confluence pages to AsciiDoc
+dtcw exportConfluence
+```
+
+### Working with Jira
+
+```bash
+# Export issues referenced in documentation
+dtcw exportJiraIssues
+
+# Export sprint changelog
+dtcw exportJiraSprintChangelog
+```
+
+### Full site generation
+
+```bash
+# Generate complete microsite
+dtcw generateSite
+
+# Preview in browser
+dtcw previewSite
+```
+
+### Docker Compose workflow (validate-first)
+
+When a docker-compose.yml with both adoc and doctoolchain services exists:
+
+```bash
+# Validate AsciiDoc first (via adcw if available)
+adcw validate -i Docs/main.adoc --strict
+
+# Then build (auto-detects compose environment)
+dtcw generatePDF
+dtcw generateHTML
+
+# Or force compose explicitly
+dtcw compose generateHTML
+```
+
+### Bootstrap a compose or devcontainer setup
+
+```bash
+# Generate docker-compose.yml with doctoolchain service
+dtcw compose install doctoolchain
+
+# Generate .devcontainer/devcontainer.json
+dtcw devcontainer install doctoolchain
+```
+
+## Error Handling
+
+- If `dtcw` is not found: install via Homebrew (`brew install tpo42/dtcw/dtcw`) or use `./dtcw` from project directory
+- If Java is missing: suggest `dtcw install java` (local mode) or use compose/docker
+- If docToolchain is not installed: suggest `dtcw install doctoolchain` or use compose/docker
+- Build timeouts: docToolchain builds can take 1-5 minutes — set long timeouts (600+ seconds)
+
+## Notes
+
+- All paths are relative to the project root
+- HTML output: `build/html5/`
+- PDF output: `build/pdf/`
+- Microsite output: `build/microsite/output/`
+- Configuration: `docToolchainConfig.groovy` (primary) or `Config.groovy` (legacy)
+- In compose mode, DTC_OPTS from the container environment are appended automatically
+- The wrapper is version-agnostic: works with both v3 (Gradle) and v4 (direct Groovy)
--- a/.github/workflows/dtcw-tests.yaml
+++ b/.github/workflows/dtcw-tests.yaml
@@ -48,7 +48,9 @@ jobs:
         run: |
           find . -maxdepth 1 -type d -not -name '.' -not -name 'test' -print0 | xargs -0 rm -rf --
           find . -maxdepth 1 -type f -not -name 'dtcw' -not -name 'test' -print0 | xargs -0 rm --
-      - name: Run tests
+      - name: Unit tests (sourced internals)
+        run: bash test/dtcw-unit.bash
+      - name: BATS integration tests
         run: |
           echo $BASH_VERSION
           test/bats/bin/bats --filter-tags \!e2e test
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,12 @@ src/test/docs/excel/
 src/test/docs/structurizr/
 ~$*
 
+# Editor swap/backup files
+.*.sw?
+*~
+*.bak
+*.orig
+
 # MAC DS_Store file
 .DS_Store
 
--- a/.gitignore
+++ b/.gitignore
@@ -62,7 +62,16 @@ resources/asciidoctor-reveal.js
 tmp/
 
 # Claude Code
-.claude/
+.claude/*
+# ... except the dtcw skill, which is a maintained project artifact rather than
+# local scratch state. Every path component has to be re-included explicitly:
+# git does not descend into an excluded directory, so a bare
+# '!.claude/skills/dtcw/SKILL.md' after '.claude/' would never match.
+!.claude/skills/
+.claude/skills/*
+!.claude/skills/dtcw/
+.claude/skills/dtcw/*
+!.claude/skills/dtcw/SKILL.md
 
 # PR Analysis Report
 PR-ANALYSIS.md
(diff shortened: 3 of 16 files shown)


rehsack:
Pushed an update. Two additions, both found by running the wrapper against a real project (a repo whose compose file pins `doctoolchain/doctoolchain:v3.5.0`).

### 1. Compose-spec file names are recognised

Auto-detect only looked for `docker-compose.yml` / `docker-compose.yaml`, so a project using the current spec name `compose.yaml` fell through to `local` or a cold `docker run`. The lookup now covers the full set, in the order `docker compose` itself resolves them — measured with all four files present:

```
compose.yaml compose.yml docker-compose.yaml docker-compose.yml  -> compose.yaml
             compose.yml docker-compose.yaml docker-compose.yml  -> compose.yml
                         docker-compose.yaml docker-compose.yml  -> docker-compose.yml
```

Matching that order matters: dtcw passes the file it picked via `-f`, so a different choice would silently override what plain `docker compose` would have used.

### 2. For container environments the version comes from the image

This was a genuine hole in the feature as first submitted. `build_command()` decided v3-vs-v4 from **host** state — `${DTC_HOME}/lib`, else `DTC_VERSION`, which defaults to the wrapper's own release. A container environment knows nothing of either: the host may have no installation at all, and the project pins its version in the compose file. With a `v3.5.0` image the wrapper built a v4 launcher command:

```
docker compose ... exec -T 'svc' bash -c "java ... '/opt/docToolchain/scripts/Launcher.groovy' 'generateHTML'"
```

That fails inside a 3.x image, which has neither `Launcher.groovy` nor `/opt/docToolchain/lib/*.jar`. Users would have had to set `DTC_VERSION=3.5.0` by hand — duplicating what the compose file already states.

The wrapper now reads the tag from the image reference (compose file, `devcontainer.json`, or the compose file a `dockerComposeFile` points at) and pins `DTC_VERSION` to it. An explicit `DTC_VERSION` still wins. A reference without a numeric version — no tag, `:latest`, a digest — is a **hard error** naming the reference it found and both remedies, rather than a guess that fails deep inside the container.

Also corrected an ADR-19 claim from the first push: it said container dispatch "calls `doctoolchain` inside the container, which handles v3/v4 internally". That holds for the v3 branch only — the v4 branch hard-codes in-container paths.

### Tests

The suite grew from 61 to 78 (`--filter-tags '!e2e'`). New `test/container_environments.bats` covers what the wrapper hands the runtime for both environments: dispatch, the `tasks` delegation, all four compose file names, `DTC_SERVICE`, version derivation for 3.x/4.x, the `DTC_VERSION` override, both hard-failure cases, and the TTY flag in each direction. Plus one case in `local_environment.bats` for the v3 path under an install directory containing a space.

argv is captured through a mock side effect rather than `mock_get_call_args()` — the latter flattens `"$@"` through `echo`, which would hide exactly the word-splitting several of these tests are about (there is a standing `# TODO: the mock doesn't handles quotes correctly` in `docker_environment.bats` about this).

Each new assertion was mutation-checked: the fix was reverted and the test confirmed to fail. That caught one weak test — the 3.x-dispatch case initially passed against host-based logic too, because pinning `DTC_VERSION` also repoints `DTC_HOME`. It now plants a host installation that looks like v4 *at the version the image pins*, so host state and image tag genuinely disagree.

Still 1 of 78 red: `sdk_installation.bats::specific doctoolchain version not installed`, unchanged and host-dependent — it fails identically on a pristine `main-4.x` worktree, because the test's `PATH` does not shield a real `docker` or a real `~/.doctoolchain`.

### One loose end for you to judge

The v3 container command ends in a literal `${DTC_OPTS:-}`:

```
... exec -T 'svc' bash -c "doctoolchain . generateHTML ${DTC_OPTS:-}"
```

The escaping reads as "let the container expand this", but `main()` hands the string to `bash -c`, and that shell expands it first. `DTC_OPTS` is not exported, so it resolves to empty on the host and the container never sees a placeholder. The outcome is right — the image reads `DTC_OPTS` from its own environment, which is where a compose file sets it — but the code says something it does not do. Happy to either drop the placeholder or escape it one level further so the container really does the expansion; I left the behaviour alone rather than change semantics late in review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rehsack:
Pushed again. Both additions are repairs to this branch's own doing: it introduced multi-runtime detection (`container`, `nerdctl`, `finch`, `podman`, `docker`) and made podman reachable in the `docker` environment, without teaching the surrounding code about either. Ported from the sibling wrapper in tpo42-asciidoc-container, where the same two gaps were fixed.

### Compose now follows the chosen runtime

`find_compose_cmd()` looked up compose independently of `detect_container_runner()`, docker-first. Two failure modes, both reproduced with stub runtimes on `PATH`:

| host | runtime chosen | compose used (before) |
|---|---|---|
| podman + docker | `podman` | **`docker compose`** — wrong engine |
| nerdctl only | `nerdctl` | **nothing** — lookup failed |

The first addresses a different engine than the one running the service. The second is worse than it looks: `get_available_environments()` requires both a compose file *and* a compose command, so on a nerdctl or finch host the `compose` environment silently disappeared from the list instead of reporting anything.

Now the lookup walks the same preference list and takes, per candidate, `<runtime> compose` or that runtime's *own* standalone binary — never crossing `docker-compose` over to podman. Measured after the change:

```
podman + docker           runner=podman    compose=<path>/podman compose
nerdctl only              runner=nerdctl   compose=<path>/nerdctl compose
podman w/o compose sub    runner=podman    compose=podman-compose      (docker-compose also present)
docker only               runner=docker    compose=<path>/docker compose
```

An explicitly set `DTC_CONTAINER_RUNNER` is authoritative: if it has no compose support the wrapper fails and names it, rather than quietly using another engine. That also avoids a regression for a host carrying Apple's `container` alongside docker — auto-detection just moves on to the next candidate, so the compose environment keeps working.

Note on the evidence: the runtimes were **stubs**, so this proves the pairing logic, not that a real `nerdctl compose` exists. For the record, `nerdctl` and `finch` implement compose as a native subcommand; `podman compose` (4.7+) only wraps an external provider, which is why the `podman-compose` fallback is not redundant.

### `--userns=keep-id` for rootless podman

The `docker` environment passes `-u $(id -u):$(id -g)` together with `-v ${PWD}:/project`. Rootless podman puts the invoking user at container uid 0 and maps every other uid into a subuid range, so that `-u` names a user existing nowhere inside the container and the bind mount is read-only in practice — a render task fails on its own output directory. The flag is added only when `podman info --format '{{.Host.Security.Rootless}}'` reports `true`, because rootful podman rejects it, and never for any other runtime.

This only became reachable because this branch replaced the hard-coded `docker` with `${DTC_CONTAINER_RUNNER}` in `build_command()`, hence a fixup here rather than a separate commit.

### Tests

`test/dtcw-unit.bash` is at 21 (was 15). Six new cases, all against stub runtimes so they run anywhere: compose pairing with a preferred runtime, a subcommand-only runtime, the matching-standalone fallback with the *other* runtime's binary also present, the explicit-runner failure, and `--userns=keep-id` present for rootless podman / absent for rootful podman and for docker.

Both changes were mutation-checked — reverting the compose lookup to its docker-first form and dropping the `keep-id` flag each make the suite fail.

Full BATS unchanged at 77 of 78, same host-dependent `sdk_installation.bats` case as before.

### Also refactored

The runtime preference list was duplicated as a `command -v` chain; it is now a single `DTC_CONTAINER_RUNNERS` array that both the runtime detection and the compose lookup read. ADR-19 documents the pairing rule, the rootless-podman flag, and adds `DTC_CONTAINER_RUNNER` and `DTC_VERSION` to its environment-variable table.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

rehsack:
The previous push turned *Java adopt-17 Build* red; fixed and green again. Worth recording why, because it corrects something I assumed earlier in this PR.

`default-build.yml` carries its own shellcheck step:

```yaml
find . -type f -perm 0755 -exec egrep -l '^#!(/bin/bash|/usr/bin/env bash)' {} + | xargs shellcheck
```

That has no `ignore_paths`, so it lints every executable bash script in the tree — including `test/dtcw-unit.bash`, which is `chmod +x`. I had earlier concluded that CI excludes `test/**` from shellcheck; that is true only of `dtcw-tests.yaml`, whose `ludeeus/action-shellcheck` step sets `ignore_paths: ./test/*`. The two workflows disagree, and the `default-build.yml` one is the stricter of the pair.

What tripped it: the new test helper generates stub runtimes, and wrote their bodies with single-quoted `echo` lines so that `$1` reaches the generated file literally. Intentional, but shellcheck flags it as SC2016, and shellcheck exits non-zero on info-level findings too. Replaced with a heredoc that interpolates only the two values it needs and escapes the stub's own positional parameters — no suppression comment required.

Verified by reproducing the CI invocation locally, restricted to what CI actually has checked out (`default-build.yml` does not fetch submodules, so `test/bats/**` and `test/test_helper/**` are absent there, and they carry plenty of pre-existing findings of their own):

```
./.ci.sh  ./bin/autobuildSite.bash  ./bin/doctoolchain  ./copyDocs.sh
./dtcw  ./dtcw4  ./netlifyBuildPreview.sh  ./scripts/injectAnalytics.sh
./test/dtcw-unit.bash
-> shellcheck exit 0
```

For whoever touches this next: `test/dtcw-unit.bash` is subject to `default-build.yml`'s shellcheck, unlike the `.bats` files, which are neither executable nor carry a bash shebang.

The divergent `ignore_paths` between the two workflows looks like something worth reconciling, but it is unrelated to this PR and I have not touched it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

